PR #90, opened 2022-11-30
Python API

jtwb:
**Status of this PR**

READY FOR REVIEW

This creates an API module with one public function, `Generate()`. Generate takes only plain Python data as input and returns a Python dictionary mapping output file names to config files (as strings).

See the docstring in api.py for a walkthrough. Additional examples are in api_test.py.
```
from aerleon import api

USE_MAIL_SERVER_SET = 1
mail_server_ips_set0 = ["200.1.1.4/32", "200.1.1.5/32"]
mail_server_ips_set1 = ["200.1.2.4/32", "200.1.2.5/32"]

networks = {
    "networks": {
        "RESERVED": {
            "values": [
                {
                    "address": "0.0.0.0/8",
                },
                {
                    "address": "10.0.0.0/8",
                },
            ]
        },
        "BOGON": {
            "values": [
                {
                    "address": "192.0.0.0/24",
                },
                {
                    "address": "192.0.2.0/24",
                },
            ]
        },
        "MAIL_SERVERS": {"values": []},
    }
}

if USE_MAIL_SERVER_SET == 0:
    networks["networks"]["MAIL_SERVERS"]["values"] = mail_server_ips_set0
else:
    networks["networks"]["MAIL_SERVERS"]["values"] = mail_server_ips_set1

cisco_example_policy = {
    "filename": "cisco_example_policy",
    "filters": [
        {
            "header": {
                "targets": {"cisco": "test-filter"},
                "kvs": {"comment": "Sample comment"},
            },
            "terms": [
                {
                    "name": "deny-to-reserved",
                    "destination-address": "RESERVED",
                    "action": "deny",
                },
                {
                    "name": "deny-to-bogons",
                    "destination-address": "BOGON",
                    "action": "deny",
                },
                {
                    "name": "allow-web-to-mail",
                    "destination-address": "MAIL_SERVERS",
                    "action": "accept",
                },
            ],
        }
    ],
}

definitions = naming.Naming()
definitions.ParseDefinitionsObject(networks, "")
configs = api.Generate([cisco_example_policy], definitions)
acl = configs["cisco_example_policy.acl"]
```

**Other changes in this PR:**

* Fixed a flaky Plugin test.

**Open issues:**

* api.py now has a big docstring but the Generate method needs a good one also.


aerleon/api.py
```diff
@@ -0,0 +1,352 @@
+"""Aerleon "Generate" API
+
+The Generate API provides the full policy-to-ACL capabilities of
+the aclgen command line tool. It accepts as input plain Python
+dictionaries and lists.
+
+This API provides a single method, "Generate()", that accepts a list of policies
+and an optional IP/ports definition object and will transform each policies
```
ankenyr, line 8: Is that really optional?

jtwb: It is not optional

aerleon/api.py
```diff
@@ -0,0 +1,352 @@
...
+cisco_example_policy = {
+    "filename": "cisco_example_policy",
+    "filters": [
+        {
+            "header": {
+                "targets": {"cisco": "test-filter"},
+                "kvs": {"comment": "Sample comment"},
+            },
+            "terms": [
+                {"name": "deny-to-reserved", "destination-address": "RESERVED", "action": "deny"},
```
ankenyr, line 38: For readability can we make all of these multiline like below?

aerleon/api.py
```diff
@@ -0,0 +1,352 @@
...
+Because this object is constructed in Python it can incorporate variable data. Users
+might conditionally construct the policy as part of a network automation workflow.
+
+Now the network names used in this example have to be defined. The naming definitions are
+constructed as follows. In this example we are dynamically selecting between two sets of IP
+addresses for the mail server.
+
+```
+mail_server_ips_set0 = ["200.1.1.4/32","200.1.1.5/32"]
+mail_server_ips_set1 = ["200.1.2.4/32","200.1.2.5/32"]
```
ankenyr, line 60: These are the same lists, maybe differentiate them?

aerleon/api.py
```diff
@@ -0,0 +1,352 @@
...
+                ),
+            )
+            async_results.append(async_result)
+        pool.close()
+        pool.join()
+        for result in async_results:
+            try:
+                result.get()
+            except (ACLParserError, ACLGeneratorError) as e:
+                logging.warning('\n\nerror encountered in rendering process:\n%s\n\n', e)
```
ankenyr, line 253: Why newlines in this error? We don't normally do newlines in error messages.

ankenyr, line 266: Would f strings be better here?

ankenyr (approved):

aerleon/api.py
jtwb: Yes, keeping this code identical to aclgen.py for the moment until we can consolidate the error handling for both cases, see #150 .

ankenyr (approved):

PR #141, opened 2023-01-15
[WIP] initial verification of arista generator doc:

fischa:
* inherited from `cisco.py`:
  * header options: * `mixed` -> missing * `enable_dsmo` -> missing * are they applicable for arista too?
  * term options:
    * `tcp-initial` is not listed in `cisco.py'
    * `verbose` is missing but default true -> also noverbose available but missing
    * expiration: removed dead links for date format


fischa:
Something odd is going on - the push to the same branch did not land in the PR. I've created a new PR #168 to update the generator docs.

PR #158, opened 2023-01-18
Run tests and codecov on pushes to main

jtwb:


codecov.yml
```diff
@@ -0,0 +1,2 @@
+comment:
+  require_changes: true
```
ankenyr, line 2: new line at end.

ankenyr (approved):

PR #182, opened 2023-01-20
JSON Schema support for Policy, Definition YAML files

jtwb:
This PR adds JSON Schema definition files for Policy, Policy Include, and Definition YAML files.

This first pass targets editor and validation scenarios. We would like users to be able to get immediate inline errors in their editors when viewing non-conforming YAML files. These schema files should also be useful in validation workflows, either for manual validation or CI/CD, git pre-commit, or automation scenario.

This PR does NOT integrate the JSON Schemas into the actual application. This might be an interesting topic of discussion - on one hand using the schema in the application helps ensure the schema never drifts from the application behavior, but a tight clamp on the performance characteristics of the validation code (startup and per-file) would be key.

image

**Status of this PR**

Seeking input.

I would like to look at providing some kind of editor config files (e.g. settings.json or example.settings.json) that would help users to map these schemas to their input files correctly. The mapping problem might require some discussion - on one hand we could re-introduce multi-part names like ".policy.yaml" so that VSCode (specifically [this YAML extension](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml)) can pick up mappings directly from [JSON Schema Store](https://www.schemastore.org/json/) . Alternatively just including the files in the repo with an example file might be pretty good.

### DIFF
--- a/schemas/aerleon-definition-yaml.schema.json
+++ b/schemas/aerleon-definition-yaml.schema.json
@@ -1,7 +1,8 @@
 {
   "$id": "urn:aerleon:schemas:aerleon-definition-yaml:1.0.0",
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "title": "AerleonDefinitionFile2",
+  "title": "AerleonDefinitionFile",
+  "description": "An Aerleon definition file defines named networks, services, or both.",
   "type": "object",
   "additionalProperties": false,
   "properties": {
--- a/schemas/aerleon-definition-yaml.schema.json
+++ b/schemas/aerleon-definition-yaml.schema.json
@@ -1,0 +1,95 @@
+{
+  "$id": "urn:aerleon:schemas:aerleon-definition-yaml:1.0.0",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "AerleonDefinitionFile2",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "networks": { "$ref": "#/$defs/networkDefinitionMapping" },
+    "services": { "$ref": "#/$defs/serviceDefinitionMapping" }
+  },
+  "$defs": {
+    "networkDefinitionMapping": {
+      "type": "object",
+      "title": "NetworkDefinitionMapping",
+      "additionalProperties": { "$ref": "#/$defs/networkDefinition" }
+    },
+    "networkDefinition": {
+      "type": "object",
+      "title": "NetworkDefinition",
+      "additionalProperties": false,
+      "required": ["values"],
+      "properties": {
+        "values": {
+          "type": "array",
+          "items": {
+            "oneOf": [
+              {
+                "type": "string",
+                "title": "NetworkCompactEntry"
+              },
+              {
+                "type": "object",
+                "title": "NetworkValueEntry",
+                "required": ["address"],
+                "properties": {
+                  "address": { "type": "string" },
+                  "comment": { "type": "string" }
+                },
+                "additionalProperties": false
+              },
+              {
+                "type": "object",
+                "title": "NetworkReferenceEntry",
+                "required": ["name"],
+                "properties": {
+                  "name": { "type": "string" },
+                  "comment": { "type": "string" }
+                },
+                "additionalProperties": false
+              }
+            ]
+          }
+        }
+      }
+    },
+    "serviceDefinitionMapping": {
+      "type": "object",
+      "title": "ServiceDefinitionMapping",
+      "additionalProperties": { "$ref": "#/$defs/serviceDefinition" }
+    },
+    "serviceDefinition": {
+      "type": "array",
+      "title": "ServiceDefinition",
+      "items": {
+        "oneOf": [
+          {
+            "type": "object",
+            "title": "ServiceValueEntry",
+            "required": ["port", "protocol"],
+            "properties": {
+              "port": {
+                "oneOf": [{ "type": "integer" }, { "type": "string" }]
+              },
+              "protocol": {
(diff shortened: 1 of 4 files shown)


nemith:
A standard file extension would be nice for policy files  but even documenting how to add a yaml.schemas and or `# yaml-language-server: $schema=https://raw.githubusercontent.com/aerleon/aerleon/master/schemas/aerleon-definition-yaml.schema.json` to a file may be good as well.

I wonder if `policy.yaml` is too generic though.  Maybe not?

jtwb:
I'm beginning to think that JSON Schema Store with a positional selector is the way to go. I poked around at the current state of the tooling ecosystem and I found that [PyCharm searches JSON Schema Store](https://www.jetbrains.com/help/pycharm/json.html#ws_json_using_schemas) and there is even a VIM extension [^1] that will do so. GitHub registers schemas for its [various](https://json.schemastore.org/github-workflow.json) [YAML](https://json.schemastore.org/github-action.json) [file formats](https://json.schemastore.org/github-issue-config.json) in JSON Schema Store.

What I like about Schema Store is it has pretty flexible targeting rules, for example we could have a rule that says any `*.yaml/*.yml` file appearing in the `policies/` directory is a Policy or Policy Include file, and files in `defs/` are Network and Service Definition files. It might be a little presumptuous of us to claim a very generic path like "defs", but I suspect we have users who keep their definition files in that path so it seems worth a shot.

I did look at the modeline option. It seems to only apply to the VSCode YAML extension and the associated `yaml-language-server` project. But it's otherwise very flexible and it might be a good Plan B.

[^1]: https://github.com/Quramy/vison

jtwb:
Quick status update: the JSON Schema files in this PR are decent and it might even make sense to merge as is. The ultimate goal of this work will be getting our schemas merged into JSON Schema store, and to do that there are guidelines they have published in terms of schema quality that I would like to implement before submitting to them.

jtwb:
https://github.com/SchemaStore/schemastore/blob/master/CONTRIBUTING.md

jtwb:
The following VSCode setting is how I configured these schemas. It might make sense to place this in some kind of example VSCode settings.json file for the time being (until we can submit to JSON Schema Store).

```
    "yaml.schemas": {
    
        "schemas/aerleon-policy-yaml.schema.json": ["*.policy.yaml", "policies/**/*.yaml"],
        "schemas/aerleon-definition-yaml.schema.json": ["*.definition.yaml", "def/**/*.yaml"],
    },
```

schemas/aerleon-definition-yaml.schema.json
```diff
@@ -0,0 +1,114 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
ankenyr, line 2: No $Id in this one?

schemas/aerleon-definition-yaml.schema.json
```diff
@@ -0,0 +1,114 @@
...
+  "$comment": "https://aerleon.readthedocs.io/en/latest/reference/naming/",
+  "title": "Aerleon Definition File",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "networks": {
+      "description": "An Aerleon definition file defines named networks and services which can be referenced by Aerleon Policy files. Networks are lists of IP addresses or CIDR IP address ranges. Services are lists of port/protocol pairs (e.g. tcp/80).\nDefinition files must include a \"networks\" section, a \"services\" section, or both.\nNetworks and services may use composition in their definitions. Include another definition by creating an item containing the \"name\" property.",
```
ankenyr, line 9: Services can also be ranges of port numbers.

ankenyr, line 9: Isn't this the help section for networks? Should it talk about services here?

schemas/aerleon-definition-yaml.schema.json
```diff
@@ -0,0 +1,114 @@
...
+    "token": {
+      "type": "string",
+      "pattern": "^[-_A-Z0-9]+$"
+    },
+    "comment": {
+      "description": "When inserting a network or service value you can attach a comment directly to that value. The comment will be loaded and potentially included in generated output on platforms that support it.",
+      "type": "string"
+    },
+    "address": {
+      "description": "When defining a network you can insert data as a specific IP address or a CIDR IP address range expression.",
```
ankenyr, line 31: Can't this also be a reference to another network definition?

jtwb: That would need to go in the `name:` field, and this is a description just for the `address:` field. The description one level above, on `values:`, does explain that you can have both in your network definition.

schemas/aerleon-definition-yaml.schema.json
jtwb: Correct, and the standard style would call for this text only discussing networks. In this version of the file I tried to use a different style where all top-level keys have copies of the same overview text. But that is not really in line with the standard style so I will change it in a future PR.

jtwb: Fixed

schemas/aerleon-definition-yaml.schema.json
jtwb: No, and I don't think $id is important for the way we are using the file right now. I will give this file another pass to make it consistent with the others, either by adding $id or removing it from all of them.

ankenyr (approved):

jtwb:
Update: Aerleon's schemas are now included in [schemastore.org](https://schemastore.org)'s catalog. VSCode users (and users of many supporting IDEs) will automatically pick up these schemas and apply them to Aerleon config, policy, and definition files.

PR #184, opened 2023-01-20
ciscoasa: Skip policy if not ASA target

btriller:


ankenyr:
Thank you for the PR @btriller
@jtwb tests in this case are not a blocker. We should probably move this logic eventually into `aclgenerator.py` or elsewhere since it is pretty universal.

ankenyr (approved):

PR #185, opened 2023-01-20
(re)format all code with black/isort

nemith:
> "Any customer can have a car painted any color that he wants so long as it
> is black".

Auto formatting source code is great for working in larger teams and allows
code review to focus on teh code and not formatting issues.

The defacto tool is "black" which is an opinonated formatting tool.  There
already seems to be configuration for black in the pyproject.toml file and
that wasn't modified.

https://black.readthedocs.io/en/stable/

In addition to "black" there is a tool called "isort" that sorts and formats
`import` lines again in the name of having a standard across all files.

https://pycqa.github.io/isort/

This PR is just running `black` and `isort` on the files and then adds a
formatting check to the github actions.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -1,4 +1,4 @@
-name: test
+name: validate
 on:
   pull_request:
     paths-ignore:
@@ -83,3 +83,10 @@ jobs:
           flags: tests
           name: codecov-umbrella
           verbose: true
+
+  format:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: psf/black@stable
+      - uses: isort/isort-action@v1.0.0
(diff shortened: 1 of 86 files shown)


jtwb:
@nemith We have a .git-blame-ignore-revs file we can use to hide large scale formatting changes from git blame. But in order to use it all of your formatting changes should be isolated to a single commit. Would you mind splitting up these changes so there is one commit with only formatting changes? All config changes should be outside of that commit.

jtwb:
We did format the code base with black not long ago in f8937ef224b027ef20bd915f35bdb02312e8d6e5. It strikes me as odd that files that have not changed since then have changes in this PR. Maybe the settings given to black were not the same?

nemith:
Black settings should be coming from pyproject.toml.   So i don't think so.

nemith:
Since i am a dummy and created this PR against my main branch on my I am going to close this and open a new one.

PR #187, opened 2023-01-20
gh workflow to validate formatting

nemith:
- (re)format all code with black/isort
- add gh workflow to check for proper code formatting (+ docs)


.github/workflows/validate.yml
```diff
@@ -83,3 +83,10 @@ jobs:
           flags: tests
           name: codecov-umbrella
           verbose: true
+
+  format:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: psf/black@stable
+      - uses: isort/isort-action@v1.0.0
```
ankenyr, line 92: nit: new line at the end of this file.

.github/workflows/validate.yml
```diff
@@ -83,3 +83,10 @@ jobs:
           flags: tests
           name: codecov-umbrella
           verbose: true
+
+  format:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: psf/black@stable
```
ankenyr, line 91: We include black in our pyproject.toml
https://github.com/aerleon/aerleon/blob/main/pyproject.toml#LL38C21-L38C21
Would using `@stable` here and `^22.8.0` in the toml file potentially cause format flip flops in the future?

jtwb: Installing black through poetry as on line 64 might be a quick way to keep things consistent:
```
      - name: Install library
        run: poetry install --no-interaction
```

jtwb: `isort` should be added as a developer dependency in `pyproject.toml` so that poetry will install both dependencies.

https://github.com/aerleon/aerleon/blob/main/pyproject.toml#L38-L53

nemith: I don't do a lot of python programming these days.  I am not familiar with poetry.  Let me look into it and perhaps we just run from poetry instead of the black action?

CONTRIBUTING.md
```diff
@@ -28,6 +28,13 @@ Writing code can include adding new features or fixing issues that have been rep
 ## Code Style and Formatting
 We have adopted [Black](https://github.com/ambv/black) as our code formatter. This takes a lot of guesswork out of formatting our code. The code we forked from adopted the [Google Python Style Guide](https://google.github.io/styleguide/pyguide.html) which we still adopt where it does not conflict with Black. Generally this means following the [Language Rules](https://google.github.io/styleguide/pyguide.html#s2-python-language-rules) while ignoring the [Style Rules](https://google.github.io/styleguide/pyguide.html#s3-python-style-rules)
 
+In additiona to Black [isort](https://pycqa.github.io/isort/) as been added to sort and format import lines.
```
jtwb, line 31:
```suggestion
In addition to Black [isort](https://pycqa.github.io/isort/) has been added to sort and format import lines.
```

jtwb:
@ankenyr Ready for review

noxfile.py
```diff
@@ -87,9 +87,10 @@ def benchmark_tuned(session):
 
 @session
 def format(session):
-    """Runs black"""
+    """Runs black and isort"""
     session.run_always("poetry", "install", external=True)
     session.run("black", "aerleon", "tests")
```
nemith, line 92: why is this scoped to just these two directories?

.github/workflows/validate.yml
```diff
@@ -83,3 +83,50 @@ jobs:
...
+        if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
+        run: poetry install --no-interaction --no-root
+      #----------------------------------------------
+      # install your root project, if required
+      #----------------------------------------------
+      - name: Install library
+        run: poetry install --no-interaction
+      #----------------------------------------------
+      # run 'black', will use installed version
+      #----------------------------------------------
```
nemith, line 125: Could probably abstract this all to a composite action in the future.

noxfile.py
jtwb: Frankly this does need a little attention. Want to change it to target "."?

.github/workflows/validate.yml
jtwb: I'm quite the GitHub Action novice but if we can abstract this duplicate behavior into one place that would be great.

noxfile.py
nemith: This is probably why i had more changes since I also formatted the tools directory.    It shouldn't hurt to change this to '.'.

.github/workflows/validate.yml
ankenyr: This is something we could do in this PR or after as a cleanup. @nemith would you be willing to follow up with one? If not please file an issue so we do not forget it.

@jtwb here is documentation on composite actions
https://docs.github.com/en/actions/creating-actions/creating-a-composite-action

.github/workflows/validate.yml
ankenyr: No worries and understandable. Looks like Jason was able to fix it.

.github/workflows/validate.yml
ankenyr: This was resolved in https://github.com/aerleon/aerleon/pull/197

ankenyr (approved):

ankenyr (approved):

PR #190, opened 2023-01-20
update .git-blame-ignore-revs

nemith:
Updates the .git-blame.ignore-revs to include the formatting commit from

```
$ git blame tools/cgrep.py | grep 186d08e | wc -l
423
$ git blame --ignore-revs-file=.git-blame-ignore-revs tools/cgrep.py | grep 186d08e | wc -l
49
```

Note: there are still lines in git blame after this change?  Because new lines were added they are still included.


jtwb (approved):

jtwb:
> Note: there are still lines in git blame after this change? Because new lines were added they are still included.

That's expected, if a new line was added by the format commit it will still appear in git blame..

Issue #74, opened 2022-10-26
[LEGACY ISSUE PR307] New flag: --[no]apply_groups to enable/disable apply-groups creation for Juniper MS-MPC


ankenyr:

btriller:
Can be closed due to #192 merged?

ankenyr:
Yup!

PR #192, opened 2023-01-23
junipermsmpc: Allow generating policies w/o apply group syntax

btriller:
Add a new keyword `no-apply-groups` for `msmpc` platform to generate policies w/o `apply-groups` syntax [2].
Currently Juniper MSMPC policies are always generated with `apply-groups` syntax [1] [3].

Also change `stdout` decorator to be able to compare generator output w/ reference files, if a test case is parameterized.

This relates to #74, but taking another approach not to add a global flag, which is only relevant for one generator.

[1] https://github.com/aerleon/aerleon/blob/99c1f65dad9ae62b457dfb786d814a0c18de4d15/tests/regression/junipermsmpc/JuniperMSMPCTest.testApplicationSets.stdout.ref#L1-L43
[2] https://github.com/aerleon/aerleon/blob/24bcc23b6a3fc6910253860a2baa49a79212a917/tests/regression/junipermsmpc/JuniperMSMPCTest.testApplicationSetsWithoutApplyGroups.stdout.ref#L1-L37
[3] https://www.juniper.net/documentation/us/en/software/junos/bgp/topics/ref/statement/apply-groups.html


ankenyr (approved):

ankenyr:
generator code looks good to me, @jtwb is looking at the test harness code.

jtwb:
This is pretty good, the ref file naming scheme mirrors the way absltest.parameterized assembles test method names. LGTM.

jtwb (approved):

PR #198, opened 2023-01-24
Move platform check into ACLGenerator

### LINKED ISSUE #188: Move 'if platform in targets' checks into ACLGenerator
See #184 for an example of a generator that was missing this check. Each generator has to include this logic at the moment.

btriller:
fixes #188

### DIFF
--- a/aerleon/lib/aclgenerator.py
+++ b/aerleon/lib/aclgenerator.py
@@ -333,15 +333,12 @@ class ACLGenerator:
                             err.append(' '.join(ns))
                 if err:
                     all_err.append(
-                        ('%s contains unsupported keywords (%s) for target ' '%s in policy %s')
+                        ('%s contains unsupported keywords (%s) for target %s in policy %s')
                         % (term.name, ' '.join(err), self._PLATFORM, pol.filename)
                     )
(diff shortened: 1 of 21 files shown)


aerleon/lib/aclgenerator.py
```diff
@@ -294,53 +294,57 @@ def __init__(self, pol, exp_info):
         """Initialise an ACLGenerator.  Store policy structure for processing."""
         supported_tokens, supported_sub_tokens = self._GetSupportedTokens()
 
+        pol.filters[:] = [
+            (header, terms)
+            for (header, terms) in pol.filters
+            if (self._PLATFORM in (header, terms)[0].platforms)
```
jtwb, line 300: Can this be simplified?
```suggestion
            if (self._PLATFORM in header.platforms)
```

jtwb (approved):
This looks great. You do not need to seek re-approval if you fix the suggestion (which is optional).

ankenyr:
Hey Bastian, this is looking really good so far. I am going to leave the review to @jtwb. We have a slack channel btw if you wish to join for closer collaboration.

Issue #40, opened 2022-10-26
[LEGACY ISSUE 174] established not added when using object-groups on cisco or ciscoxr


ankenyr:

ankenyr:
This issue references https://github.com/google/capirca/issues/174

PR #199, opened 2023-01-24
Reorganizing object-group and extended term string functions.

ankenyr:
This starts the fixes for https://github.com/aerleon/aerleon/issues/40 and https://github.com/aerleon/aerleon/issues/41
We want to consolidate these so that we no longer have to worry about keeping multiple string functions in sync. Where there are differences those should be handled with inheritance or other ways to abstract the differences.

- Created a Term object for Extended and Object terms to inherit from, standard needs more work to inherit.
- StandardTerm now uses a `prefix` string to cut down on multiple if/else conditionals that are same beyond the prefix.
- Consolidated common logic of Extended and Object into the __init__ of `Term`.
- ObjectTerm has its own `_GetIpString` function as it differs from the standard.
- Sorted all lists in term constructor (line 633).
- Summarizer now has a hash function similar to IP objects in the standard library.

### DIFF
--- a/aerleon/lib/cisco.py
+++ b/aerleon/lib/cisco.py
@@ -571,7 +571,7 @@ class Term(aclgenerator.Term):
         else:
             # source address not set
             source_address = ['any']
-
+        source_address_set = set(source_address)
         # destination address
         if self.term.destination_address:
             destination_address = self.term.GetAddressOfVersion('destination_address', self.af)
@@ -594,6 +594,7 @@ class Term(aclgenerator.Term):
         else:
             # destination address not set
             destination_address = ['any']
+        destination_address_set = set(destination_address)
 
         # ports
         source_port = [()]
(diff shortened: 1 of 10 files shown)


jtwb:
Tests are failing: https://github.com/aerleon/aerleon/actions/runs/4061580663/jobs/7089463558

jtwb:
I was able to rebase this branch on top of the recent changes to cisco.py in main, and that actually fixed the tests. See branch [issue174_2_rebased](https://github.com/aerleon/aerleon/compare/issue174_2_rebased?expand=1).
```
nox > Ran multiple sessions:
nox > * test-3.7: success
nox > * test-3.8: success
nox > * test-3.9: success
nox > * test-3.10: success
```
You may want to take a look at that branch and use `git reset --hard issue174_2_rebased` / `git push --force` to update this PR.

aerleon/lib/cisco.py
```diff
@@ -475,7 +428,6 @@ def __init__(
         platform='cisco',
         verbose=True,
     ):
-        super().__init__(term)
```
jtwb, line 478: Why does `Cisco.Term.__init__` no longer call `super().__init__`? This causes UnsupportedFilterError checks to be skipped and stops ProtocolNameToNumber() from being called on term.protocol.

https://github.com/aerleon/aerleon/blob/2820a3920238460d2143224ee6d4cc8c1eae2b27/aerleon/lib/aclgenerator.py#L128-L139

ankenyr: Added this back, I believe it was giving me issues at one point.

jtwb (approved):

PR #205, opened 2023-01-26
Move expiration check to ACLGenerator

### LINKED ISSUE #33: Hoist "expiration" impl into core
Currently the implementation of the 'expiration' property is handled through a block of copy-paste code in every generator. Should a generator author forget to paste in this code, expiration won't work.

The core application should handle the expiration implementation and not even pass it through to the generator.

btriller:
fixes #33

### DIFF
--- a/aerleon/lib/aclgenerator.py
+++ b/aerleon/lib/aclgenerator.py
@@ -16,12 +16,13 @@
 
 """ACL Generator base class."""
 
-import datetime
 import copy
+import datetime
 import hashlib
-from absl import logging
 import re
 import string
+
+from absl import logging
 
 from aerleon.lib import policy
 
--- a/aerleon/lib/aclgenerator.py
+++ b/aerleon/lib/aclgenerator.py
@@ -16,9 +16,10 @@
 
 """ACL Generator base class."""
 
+import datetime
 import copy
 import hashlib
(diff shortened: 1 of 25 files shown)


aerleon/lib/aclgenerator.py
```diff
@@ -302,7 +304,34 @@ def __init__(self, pol, exp_info):
         self.policy = pol
         all_err = []
         all_warn = []
+        current_date = datetime.datetime.utcnow().date()
+        exp_info_date = current_date + datetime.timedelta(weeks=exp_info)
         for header, terms in pol.filters:
+            filter_name = header.FilterName(self._PLATFORM)
+            terms[:] = [
```
ankenyr, line 311: What made you choose to do a list comprehension in this instance? Once list comprehensions get past a one liner they tend to become harder to read.

btriller: Started out as a oneliner and grew after adding logging. Should I change to `filter()` then?

ankenyr: They always do haha
Yea a filter would be more readable here.

ankenyr (approved):

ankenyr:
@all-contributors lets add @btriller for code

PR #208, opened 2023-01-26
Fix issues with include files in YAML

jtwb:
Thanks to @nemith for raising these issues in slack.

This fixes two issues preventing users from using includes in YAML files in aerleon 1.0.1.

* Having a YAML include file in the policy directory would cause a crash with `AttributeError: 'NoneType' object has no attribute 'filters'`.
* Including a YAML include file would cause a crash with `ValueError: I/O operation on closed file.`.

This PR also updates the docs and adds an e2e test with YAML includes.


ankenyr (approved):

aerleon/lib/yaml.py
```diff
@@ -330,7 +334,7 @@ def _LoadIncludeFile(base_dir, inc_filename):
     """Open an include file."""
 
     with open(pathlib.Path(base_dir).joinpath(inc_filename), 'r') as include_file:
-        return include_file
+        return include_file.read()
```
nemith, line 337: Do you really want to load this into memory?  

You could just move `with open()` to the one spot where it is called? 

```patch
-                include_file = _LoadIncludeFile(base_dir, inc_filename)
-                include_data = yaml.load(
-                    include_file, Loader=SpanSafeYamlLoader(filename=inc_filename)
-                )
+                with open(pathlib.Path(base_dir).joinpath(inc_filename), 'r') as include_file:
+                    include_data = yaml.load(
+                        include_file, Loader=SpanSafeYamlLoader(filename=inc_filename)
+                    )
```

jtwb: Yeah, `yaml` loader would stream through the file nicely so we are balooning our process memory a bit by loading in the whole file. In my experience policy files - especially includes - do not seem to be so large that this will be noticeable to the user (unlike network/service definitions which can be truly large).

`_LoadIncludeFile` exists purely for unit test mocks, and those tests really could just mock `open` (as do many similar tests). So my suggestion would be to just do away with `_LoadIncludeFile` completely and focus on mocking `open`.

jtwb: https://github.com/aerleon/aerleon/issues/209

PR #212, opened 2023-01-29
Use ProcessPoolExecutor for multiprocessing and fix logging

nemith:
This switches to using a ProcessPoolExecutor which is a bit cleaner way
of handling a multiprocessing pool and also allows for RenderFile and
RenderACL to return it's results instead of appending to an existing
thread-safe list which is much more straight forward.

In addition logging was only being setup on the main process.  This
calls a setup_logging() function at the start of each multiprocessing
function to make sure all logging is done for all subprocesses (which
should make it match when multiprocessing isn't used).

This will allow a future (already written) PR to remove absl logging in
favor for the standar library `logging function`

### DIFF
--- a/aerleon/aclgen.py
+++ b/aerleon/aclgen.py
@@ -16,10 +16,10 @@
 """Renders policy source files into actual Access Control Lists."""
 
 import copy
-import multiprocessing
 import pathlib
 import sys
-from typing import Iterator, List, Tuple
+from concurrent.futures import Future, ProcessPoolExecutor
+from typing import Any, Callable, Iterator, List, Optional, Tuple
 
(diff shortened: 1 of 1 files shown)


ankenyr:
@all-contributors, let's add @nemith for code contributions.

ankenyr:
@all-contributors please add @nemith for code

ankenyr:
Hi Brandon, we saw this is still in draft still. Were you still working on things here?

ankenyr:
Hey @nemith just checking in again. Is this expected to be a draft or are you looking for a review?

nemith:
This is still a draft.  I was going to fix logging first, but it's already pretty broken.

PR #218, opened 2023-02-01
Move term platform check to ACLGenerator

btriller:


tests/regression/aruba/aruba_test.py
```diff
@@ -875,6 +895,39 @@ def testNegateAny(self):
...
+        self.naming.GetServiceByProto.return_value = ['69']
+        aru = aruba.Aruba(
+            policy.ParsePolicy(
+                GOOD_HEADER_V4 + GOOD_TERMS_COMBINED_SINGLE_CASE + PLATFORM_TERM, self.naming
+            ),
+            EXP_INFO,
+        )
+        output = str(aru)
+        self.assertNotIn('platform-term', output, output)
+        print(output)
```
jtwb, line 912: Not a blocker: it looks like the `@capture.stdout` decorator is missing. Including it for this test and testPlatformExclude would cause .ref files to be created for these tests.

policies/pol/sample_arista_tp.pol
```diff
@@ -23,6 +23,16 @@ term wonky-prots-except {
   action:: accept
 }
 
+term accept-traceroute-exc {
```
jtwb, line 26: There is a file, `sample_arista_tp_yaml.yaml` that contains this same example but in YAML. Can you please update that file also?

jtwb (approved):

PR #228, opened 2023-02-02
fix broken `cgrep -o <object>` lookup

fischa:
Lookup of an object with `cgrep -o <object>` did return: 
```text
  File "XXXXX/.venv/lib/python3.9/site-packages/tools/cgrep.py", line 241, in main
    ips.sort(key=lambda x: int(x.ip)) 
  File "XXXXX/.venv/lib/python3.9/site-packages/tools/cgrep.py", line 241, in <lambda> 
    ips.sort(key=lambda x: int(x.ip)) 
AttributeError: 'IPv4' object has no attribute 'ip' 
```
With the suggest fix it works as expected.

### DIFF
--- a/tools/cgrep.py
+++ b/tools/cgrep.py
@@ -238,8 +238,8 @@ def main(parser):
             else:
                 logging.info(token + ':')
                 # convert list of ip objects to strings and sort them
-                ips.sort(key=lambda x: int(x.ip))
-                p([str(x) for x in ips])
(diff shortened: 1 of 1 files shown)


fischa:
It was an issue in capirca too: https://github.com/google/capirca/issues/225 😄

ankenyr:
We really should get some unit tests for cgrep and aclcheck at some point if only to help us built new better versions at some point.

ankenyr (approved):

Issue #226, opened 2023-02-02
subnet comparison in aclcheck.py not working as expected


calblox:
I noticed that aclcheck didn't seem to be working as I expected- packets weren't matching terms I would have expected them to.

After a bit of debugging, I isolated this line as what I believe to be the root cause:
https://github.com/aerleon/aerleon/blob/61c0bc93a1366872d6798d2ca6d02509b8b5fe49/aerleon/lib/aclcheck.py#L262

I think this would work if nacaddr had the in operator overloaded to support comparisons like this, and I suspect that may be the original intent (similar to ipaddr?).

I was able to workaround the issue by changing this evaluation to:
```if addr.subnet_of(next):```, and verified that this worked as expected.

I think subnet_of also handles version comparison, so line 261 can probably be pruned as superfluous if this fix is adopted:
https://github.com/aerleon/aerleon/blob/61c0bc93a1366872d6798d2ca6d02509b8b5fe49/aerleon/lib/aclcheck.py#L261

Steps to repro issue using an included pol file- I believe both of these results are incorrect:
```
(aerleon-py3.11) cwarlick@HQ-C02ZL1ZGLVDR aerleon % python ./aerleon/aclcheck_cmdline.py -s '10.0.0.10' -d '10.0.0.11' --proto=udp --sport=53 -p policies/pol/sample_juniper_loopback.pol
  filter: LOOPBACK
          term: discard-default
                deny
(aerleon-py3.11) cwarlick@HQ-C02ZL1ZGLVDR aerleon % python ./aerleon/aclcheck_cmdline.py -s '123.123.123.123' -d '10.0.0.11' --proto=udp --sport=53 -p policies/pol/sample_juniper_loopback.pol
  filter: LOOPBACK
          term: discard-default
                deny

(aerleon-py3.11) cwarlick@HQ-C02ZL1ZGLVDR aerleon % python3
Python 3.11.1 (main, Dec 23 2022, 09:40:52) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from aerleon.lib import nacaddr
>>> dot0_slash31 = nacaddr.IP("10.255.255.0/31")
>>> dot1_slash32 = nacaddr.IP("10.255.255.1/32")
>>> dot2_slash32 = nacaddr.IP("10.255.255.2/32")
>>> dot1_slash32.subnet_of(dot0_slash31)
True
[...]

PR #229, opened 2023-02-02
subnet comparison in aclcheck.py not working as expected

calblox:
possible fix addressing problem described in https://github.com/aerleon/aerleon/issues/226

tests run clean with one seemingly unrelated warning-
====== 1882 passed, 1 warning in 40.79s 

 in tests/utils/iputils_test.py:28- Unknown pytest.mark.unit

### DIFF
--- a/aerleon/lib/aclcheck.py
+++ b/aerleon/lib/aclcheck.py
@@ -257,10 +257,8 @@ class AclCheck(object):
         if not addresses:
             return True  # always true if term has nothing to match
         for next in addresses:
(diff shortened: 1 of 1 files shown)


ankenyr:
This looks good, a good followup would be to make some tests for aclcheck. I believe there are a bunch of edge cases that are not working and that would be impactful work. Let me know if you wish to tackle that yourself for the project else we can handle it at some point.

ankenyr (approved):

PR #238, opened 2023-02-04
Migrate to Material theme docs

itdependsnetworks:
I think the Material theme looks a bit more modern. I know there was some confusion around the theme last time and how it interacts with RTD, but it is just a theme, sphinx just happens to be the default for rtd but not enforced.

Additionally enforced https://github.com/DavidAnson/markdownlint/blob/main/doc/md025.md standard.

image

### DIFF
--- a/docs/reference/yaml_reference.md
+++ b/docs/reference/yaml_reference.md
@@ -247,7 +247,7 @@ YAML supports in-file code reuse through aliases and anchors. This is not an
 Aerleon-specific feature; YAML automatically resolves aliases as part of the
 loading process.
 
-# YAML Policy File Spec
+## YAML Policy File Spec
 
 | File Structure          |                                                           |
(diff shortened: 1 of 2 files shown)


ankenyr:
Hey @jtwb have you given any more thought to this since we last spoke about it?

jtwb:
> Hey @jtwb have you given any more thought to this since we last spoke about it?

I left some comments about stylistic and navigation issues in Slack here: https://aerleon.slack.com/archives/C04KT4BJ7J9/p1675917317529349

ankenyr:
Just to bring those comments into here because they will eventually get lost in slack
1) Reference section seems to be collapsed
2) Some code examples are smaller font sizes than others when they are in expandable code blocks
3) The copy button on code blocks looks to cover some code if it only has a few lines.

jtwb (approved):

PR #244, opened 2023-02-09
Add support for aerleon.yml file

### LINKED ISSUE #237: Config file improvements
At the moment, `aclgen` can read configuration settings like `base_directory`, `definitions_directory` , `exp_info` and others from a YAML config file specified with --config_file.

image

### 1. `aclcheck` should read the same config file.

It would be convenient to enable `aclcheck` to also read from the same config file. `base_directory` and `definitions_directory` specifically are likely going to be the same across both tools, so it would be convenient for the user to just use the same config for both.

`pol2yaml` is headed in the same direction, as it also needs `base_directory`, etc.

### 2. The config file should have a standard file name and location.

Currently the config file can be named anything so long as it is YAML. IMO it would be more standard and ergonomic to have a default name and location for the config YAML file, e.g. `aerleon.yml`. A standard name and location would match familiar python tools and it would let us set a schema on the file through JSON Schema, so users can get auto-completion and error checking in their editor.

### 3. This would still work if the CLI tools were unified.

These changes would anticipate a unified CLI if we go that route in the future.

### Proposal:

1. `aclgen`, `aclcheck`, and `pol2yaml` should check for the existence of a file named `aerleon.yml` in the current directory. If present this file will be loaded as a config file.
2. The user can still set --config_file to select a different config file.
3. The user can still override these defaults using command line args (e.g. --base_directory).
4. A JSON Schema will be registered for `aerleon.yml` so users get auto-completion and error checking in-editor (VSCode, PyCharm, VIM).

jtwb:
Fixes #237 

image

### Changes in this PR: aerleon.yml

* `aclgen` and `aclcheck` will now try to read configuration settings from `aerleon.yml` in the current directory.
* Configuration settings found in `aerleon.yml` can be overridden by command line arguments.
* `aclgen` and `aclcheck` can be given the command line argument `--config_file` to use a different config file instead. In the `aclgen` case it is still possible to provide multiple config files with this argument (this was supported before this PR).
* It is not an error if `aerleon.yml` is not found.
* An example `aerleon.yml` file is included in this repo with some default settings.

### Other changes in this PR

* `aclgen` now gives command line arguments precedence over settings found in config files.
* `aclcheck` is upgraded from `optparse` to `argparse`.
* All `absl`-specific code was removed from `aerleon.utils.config`. The config module interface is simplified.

### Future changes

* We should add a JSON Schema for `aerleon.yml` for in-editor validation, autocomplete, and tooltips.

### DIFF
--- a/aerleon/aclcheck_cmdline.py
+++ b/aerleon/aclcheck_cmdline.py
@@ -17,72 +17,105 @@
 """Command line interface to aclcheck library."""
 
 import pathlib
-from optparse import OptionParser
+from argparse import ArgumentParser, RawTextHelpFormatter
 
 from aerleon.lib import aclcheck, naming, policy, yaml
+from aerleon.utils import config
 
 
 def main():
     # TODO(robankeny): Lets move this to gflags
-    usage = 'usage: %prog [options] arg'
-    _parser = OptionParser(usage)
-    _parser.add_option(
+    # usage = 'usage: %prog [options] arg'
+    _parser = ArgumentParser(
+        prog='aclcheck',
+        formatter_class=RawTextHelpFormatter,
+    )
+    _parser.add_argument(
         '--definitions-directory',
-        dest='definitions',
+        dest='definitions_directory',
         help='definitions directory',
-        default='./def',
     )
-    _parser.add_option(
+    _parser.add_argument(
         '--base-directory',
-        dest='basedir',
+        dest='base_directory',
         help='The base directory to look for include files.',
-        default='',
     )
-    _parser.add_option(
+    _parser.add_argument(
         '-p',
         '--policy-file',
         dest='pol',
         help='policy file',
-        default='./policies/pol/sample_paloalto_yaml.yaml',
+        required=True,
     )
-    _parser.add_option(
-        '-d', '--destination', dest='dst', help='destination IP', default='200.1.1.1'
+    _parser.add_argument(
+        '--config-file',
+        dest='config_file',
+        help='config file',
     )
-    _parser.add_option('-s', '--source', dest='src', help='source IP', default='any')
-    _parser.add_option(
+    _parser.add_argument('-d', '--destination', dest='dst', help='destination IP')
+    _parser.add_argument(
+        '-s',
+        '--source',
+        dest='src',
+        help='source IP',
+    )
+    _parser.add_argument(
         '--proto',
         '--protocol',
         dest='proto',
         help='Protocol (tcp, udp, icmp, etc.)',
-        default='tcp',
     )
-    _parser.add_option(
-        '--dport', '--destination-port', dest='dport', help='destination port', default='80'
-    )
-    _parser.add_option(
-        '--sport', '--source-port', dest='sport', help='source port', default='1025'
-    )
-    (FLAGS, unused_args) = _parser.parse_args()
+    _parser.add_argument('--dport', '--destination-port', dest='dport', help='destination port')
+    _parser.add_argument('--sport', '--source-port', dest='sport', help='source port')
+    FLAGS = _parser.parse_args()
 
-    defs = naming.Naming(FLAGS.definitions)
-    with open(FLAGS.pol) as f:
+    default_flags = {
+        'base_directory': './policies',
+        'definitions_directory': './def',
+        'pol': None,
+        'config_file': None,
+        'dst': '200.1.1.1',
+        'src': 'any',
+        'proto': 'any',
+        'dport': '80',
+        'sport': '1025',
+    }
+
+    configs = {}
+    configs.update(default_flags)
+
+    if not (FLAGS.base_directory and FLAGS.definitions_directory):
+        common_flags = frozenset(['base_directory', 'definitions_directory'])
+        common_configs = config.load_config(config_file=FLAGS.config_file)
+        common_configs = {
+            key: value for key, value in common_configs.items() if key in common_flags
+        }
+        configs.update(common_configs)
+
+    # Grab common configs from aerleon.yml
+    configs.update({flag: value for flag, value in vars(FLAGS).items() if value})
+
+    defs = naming.Naming(configs['definitions_directory'])
+
+    with open(configs['pol']) as f:
         conf = f.read()
-    if pathlib.Path(FLAGS.pol).suffix in ['.yaml', '.yml']:
+
+    if pathlib.Path(configs['pol']).suffix in ['.yaml', '.yml']:
         policy_obj = yaml.ParsePolicy(
             conf,
-            base_dir=FLAGS.basedir,
-            filename=FLAGS.pol,
+            base_dir=configs['base_directory'],
+            filename=configs['pol'],
             definitions=defs,
         )
     else:
-        policy_obj = policy.ParsePolicy(conf, defs, base_dir=FLAGS.basedir)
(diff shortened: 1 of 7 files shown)


aerleon/aclcheck_cmdline.py
```diff
@@ -17,72 +17,104 @@
...
+    _parser.add_argument('--dport', '--destination-port', dest='dport', help='destination port')
+    _parser.add_argument('--sport', '--source-port', dest='sport', help='source port')
+    FLAGS = _parser.parse_args()
+
+    default_flags = {
+        'base_directory': './policies',
+        'definitions_directory': './def',
+        'pol': None,
+        'config_file': None,
+        'dst': '200.1.1.1',
```
ankenyr, line 76: It is strange to me that this was selected as a default.

aerleon/aclcheck_cmdline.py
```diff
@@ -17,72 +17,104 @@
...
+        'src': 'any',
+        'proto': 'any',
+        'dport': '80',
```
ankenyr, line 79: Same for these.

aerleon/utils/config.py
```diff
@@ -19,55 +21,49 @@
...
-    }
+    Args:
+        config_file: An optional string or pathlib.Path with the location of the config file
+            to open. A list of config files can be given. Default is './aerleon.yml'.
+        apply_defaults: Whether to set missing config fields to their default values. Default
+            is True.
 
-    return {flag: base[flag] for flag in filter(lambda f: base[f] is not None, base)}
+    Raises:
+        IOError: If open(config_file) raises an IOError. Note that it is not an error for the
```
ankenyr, line 40: Don't you mean FileNotFoundError?

aerleon/aclcheck_cmdline.py
```diff
@@ -17,72 +17,104 @@
...
+    }
 
-    defs = naming.Naming(FLAGS.definitions)
-    with open(FLAGS.pol) as f:
+    configs = {}
+    configs.update(default_flags)
+
+    if not (FLAGS.base_directory and FLAGS.definitions_directory):
+        common_flags = frozenset(['base_directory', 'definitions_directory'])
+        common_configs = config.load_config(config_file=FLAGS.config_file)
```
ankenyr, line 88: Shouldn't you be handling the file not existing error here?

aerleon/aclcheck_cmdline.py
```diff
@@ -17,72 +17,104 @@
...
+            key: value for key, value in common_configs.items() if key in common_flags
+        }
+        configs.update(common_configs)
+
+    # Grab common configs from aerleon.yml
+    configs.update({flag: value for flag, value in vars(FLAGS).items() if value})
+
+    defs = naming.Naming(configs['definitions_directory'])
+
+    with open(configs['pol']) as f:
```
ankenyr, line 99: This can raise an error when then file does not exist and should be caught providing a better error message.

aerleon/aclcheck_cmdline.py
```diff
@@ -17,72 +17,104 @@
...
+        'dst': '200.1.1.1',
+        'src': 'any',
+        'proto': 'any',
+        'dport': '80',
+        'sport': '1025',
+    }
 
-    defs = naming.Naming(FLAGS.definitions)
-    with open(FLAGS.pol) as f:
+    configs = {}
```
ankenyr, line 83: Can we choose a different name? It makes the following section harder to read having configs as a dict and config as a imported library. I would suggest options or even conf. Anything but a single letter difference.

aerleon/aclcheck_cmdline.py
```diff
@@ -17,72 +17,104 @@
...
+    configs.update(default_flags)
+
+    if not (FLAGS.base_directory and FLAGS.definitions_directory):
```
ankenyr, line 86: If I am understanding this section correctly you are merging the command line flags onto the ones retrieved from the config file? But only in the case where base_directory and definition_directory are not set? Why not merge it always?

aerleon/aclcheck_cmdline.py
jtwb: It's some random IP in LACNIC... very odd indeed.

aerleon/utils/config.py
jtwb: No, the docstring is correct. FileNotFoundError is handed in this function on line 66. The desired behavior is for load_config to not raise an error if there is not config file. IOError can also be raised by `open()` if there is some more subtle issue with the filesystem and in that case it is up to the caller to decide what to do.

From the docstring on line 44/45:

```
Returns: a dictionary containing the contents of the config file. If no config file is found
        all default values will be returned (unless apply_defaults is False).
```

aerleon/aclcheck_cmdline.py
jtwb: aclcheck only uses the config file fields base_directory and definitions_directory. If the user provided both options on the command line we can skip reading the config file since the command line values take higher precedence.

aerleon/utils/config.py
jtwb: Also from lines 40/41:

```
Note that it is not an error for the config file to not exist.
```

aerleon/aclcheck_cmdline.py
jtwb: No, see the docstring for config.load_config . FileNotFoundError is handled inside load_config as it is not an error if there is no aerleon.yml file (which will be the case for most users).

If opening the config file fails with IOError we do want to terminate the process and inform the user. In that case it might be a little better to catch and format the error.

https://peps.python.org/pep-3151/#common-errnos-with-ioerror

#### Common errnos with IOError

- EACCES: permission denied (for filesystem operations)
- EBADF: bad file descriptor (with select.epoll); read operation on a write-only GzipFile, or vice-versa
- EBUSY: device or resource busy
- EISDIR: is a directory (when trying to open())
- ENODEV: no such device
- ENOENT: no such file or directory (when trying to open())
- ETIMEDOUT: connection timed out

aerleon/aclcheck_cmdline.py
jtwb: I don't want to expand the scope of this PR to include selecting new default values for aclcheck. Odd as they are. Will file an issue.

aerleon/aclcheck_cmdline.py
jtwb: See previous comment. Changing the default values is not in scope for this PR.

aerleon/aclcheck_cmdline.py
ankenyr: Yea sorry, didn't mean to suggest that. More asking for your opinions on the oddness.

aerleon/utils/config.py
```diff
@@ -19,55 +21,49 @@
     'exp_info': 2,
 }
 
+default_file = './aerleon.yml'
```
ankenyr, line 24: https://peps.python.org/pep-0008/#constants
Surprised none of our stuff caught that this is a constant?

ankenyr (approved):

aerleon/utils/config.py
jtwb: I've fixed `config.DEFAULT_FILE` but let `config.defaults` remain incorrect for the moment. I'll put that in the follow up PR.

aerleon/aclcheck_cmdline.py
jtwb: I've updated the error handling behavior per slack convo.

PR #248, opened 2023-02-12
Add api.AclCheck and accept PolicyDict input in more places

jtwb:
Status of this PR:

* Ready to review

Changes in this PR

* Adds `api.AclCheck()` which accepts a policy dictionary and returns a list of matches grouped by filter, term. This mirrors the command-line aclcheck functionality.
* Defines a type 'PolicyDict' for the existing policy dictionary used by the API.
* PolicyBuilder can now accept a PolicyDict, so users can create a Policy model directly from a dictionary. RawPolicy is no longer part of the public interface.
* The YAML policy front-end no longer needs to generate a RawPolicy so the code has been simplified.

Other changes

* Test coverage for AclCheck was expanded.
* AclCheck.\_\_str\_\_ (used by the command line) and AclCheck.Summarize (used by the API) share the same code.
* AclCheck has a new constructor FromPolicyDict so users can create an AclCheck instance directly from PolicyDict for advanced usage.

### DIFF
--- a/aerleon/lib/policy_builder.py
+++ b/aerleon/lib/policy_builder.py
@@ -34,8 +34,124 @@ if sys.version_info < (3, 9):
 else:
     from typing import Annotated
 
+if sys.version_info < (3, 10):
+    from typing_extensions import TypeAlias
+else:
+    from typing import TypeAlias
+
+if sys.version_info < (3, 11):
+    from typing_extensions import NotRequired, Required
+else:
+    from typing import NotRequired, Required
+
+if sys.version_info < (3, 8):
+    from typing_extensions import TypedDict
+else:
+    from typing import TypedDict
+
 if typing.TYPE_CHECKING:
+    from datetime import datetime
+
     from aerleon.lib import naming
+
+
+# NOTE: TypedDict is just a normal dictionary, it is not a class. This syntax tells the type checker
+# what keys need to be present in the given dictionary.
+
+WordList: TypeAlias = "str | list[str]"
+
+VPNValue = TypedDict('VPNValue', {"name": str, "policy": "NotRequired[str]"})
+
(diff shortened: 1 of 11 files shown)


aerleon/lib/policy_builder.py
```diff
@@ -34,10 +34,126 @@
...
+    from typing_extensions import TypeAlias
+else:
+    from typing import TypeAlias
+
+if sys.version_info < (3, 11):
+    from typing_extensions import NotRequired, Required
+else:
+    from typing import NotRequired, Required
+
+if sys.version_info < (3, 8):
```
ankenyr, line 47: Can order these by their versions?

jtwb: There is probably a tidy way to express this kind of thing. But ordering by version would be an improvement.

tests/lib/yaml_test.py
```diff
@@ -252,6 +252,18 @@ def testTypeErrors(self):
             "Filter must contain a terms section. File=policy_no_terms.pol.yaml, Line=3.",
         )
 
+        with self.assertRaises(yaml_frontend.PolicyTypeError) as arcm:
+            yaml_frontend.ParsePolicy(
+                IGNORED_YAML_POLICY_NO_TERMS,
+                filename="policy_no_targets.pol.yaml",
```
ankenyr, line 258: do we want to remove `.pol.yaml` and use just `.yaml`

jtwb: It looks like this file contains a few dozen references to ".pol.yaml". Probably better to call this out of scope and do a follow up PR to avoid making this diff any larger.

ankenyr: sgtm

ankenyr (approved):

PR #250, opened 2023-02-14
Adding typing to nacaddr, naming, and iputils using monkeytype.

ankenyr:
This PR was created using monkeytype against tests. An example of this would be 
```
poetry run monkeytype run tests/lib/nacaddr_test.py
```
This runs the test and similar to a profiler, hooks into method calls figuring out what types are being used in the functions.

### DIFF
--- a/aerleon/lib/nacaddr.py
+++ b/aerleon/lib/nacaddr.py
@@ -16,15 +16,22 @@
 
 """A subclass of the ipaddress library that includes comments for ipaddress."""
 
+from __future__ import annotations
+
 import collections
 import ipaddress
 import itertools
-from typing import Union
+from typing import Any, DefaultDict, List, Optional, Tuple, Union
 
 import aerleon.utils.iputils as iputils
 
 
-def IP(ip, comment='', token='', strict=True):
+def IP(
+    ip: Union[ipaddress.IPv4Network, ipaddress.IPv6Network, str],
+    comment: str = '',
+    token: str = '',
+    strict: bool = True,
+) -> Union[IPv4, IPv6]:
     """Take an ip string and return an object of the correct type.
 
     Args:
@@ -52,7 +59,9 @@ def IP(ip, comment='', token='', strict=True):
 
 # TODO(robankeny) remove once at 3.7
 @staticmethod
-def _is_subnet_of(a, b):  # pylint: disable=invalid-name
+def _is_subnet_of(
+    a: Union[IPv4, IPv6], b: Union[IPv4, IPv6]
+) -> bool:  # pylint: disable=invalid-name
     try:
         # Always false if one is v4 and the other is v6.
         if a.version != b.version:
@@ -67,7 +76,13 @@ def _is_subnet_of(a, b):  # pylint: disable=invalid-name
 class IPv4(ipaddress.IPv4Network):
     """This subclass allows us to keep text comments related to each object."""
 
-    def __init__(self, ip_string, comment='', token='', strict=True):
+    def __init__(
+        self,
+        ip_string: Union[ipaddress.IPv4Network, Tuple[int, int], str, IPv4],
+        comment: Union[str, IPv4] = '',
+        token: str = '',
+        strict: bool = True,
+    ) -> None:
         self.text = comment
         self.token = token
         self.parent_token = token
@@ -83,13 +98,13 @@ class IPv4(ipaddress.IPv4Network):
             ip = ip_string
         super().__init__(ip, strict)
 
-    def subnet_of(self, other):
+    def subnet_of(self, other: "IPv4") -> bool:
         """Return True if this network is a subnet of other."""
         if self.version != other.version:
             return False
         return self._is_subnet_of(self, other)
 
-    def supernet_of(self, other):
+    def supernet_of(self, other: Union[IPv4, IPv6]) -> bool:
         """Return True if this network is a supernet of other."""
         if self.version != other.version:
             return False
@@ -102,7 +117,7 @@ class IPv4(ipaddress.IPv4Network):
         result.parent_token = self.parent_token
         return result
 
-    def AddComment(self, comment=''):
+    def AddComment(self, comment: str = '') -> None:
         """Append comment to self.text, comma separated.
 
         Don't add the comment if it's the same as self.text.
@@ -116,7 +131,7 @@ class IPv4(ipaddress.IPv4Network):
         else:
             self.text = comment
 
-    def supernet(self, prefixlen_diff=1):
+    def supernet(self, prefixlen_diff: int = 1) -> "IPv4":
         """Override ipaddress.IPv4 supernet so we can maintain comments.
 
         See ipaddress.IPv4.Supernet for complete documentation.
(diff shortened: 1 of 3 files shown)


aerleon/lib/nacaddr.py
```diff
@@ -286,7 +307,14 @@ def CollapseAddrListPreserveTokens(addresses):
     return [i for sublist in dedup_array for i in sublist]
 
 
-def _SafeToMerge(address, merge_target, check_addresses):
+def _SafeToMerge(
+    address: Union[IPv4, IPv6],
+    merge_target: Union[IPv4, IPv6],
+    check_addresses: Union[
+        DefaultDict[ipaddress.IPv4Address, List[IPv4]],
+        DefaultDict[ipaddress.IPv6Address, List[IPv6]],
```
jtwb, line 315: DefaultDict might be passed to this function in all cases, but this function doesn't use any features of DefaultDict. Better to be more generic and use Dict.

aerleon/lib/nacaddr.py
```diff
@@ -238,23 +259,23 @@ def AddComment(self, comment=''):
...
+def _InNetList(adders: List[IPv4], ip: IPv4) -> bool:
     """Returns True if ip is contained in adders."""
     for addr in adders:
         if ip.subnet_of(addr):
             return True
     return False
 
 
-def IsSuperNet(supernets, subnets):
+def IsSuperNet(supernets: List[IPv4], subnets: List[IPv4]) -> bool:
```
jtwb, line 270: Is this code actually restricted to IPv4 only? IT just calls `_InNetList` which calls `ip.subnet_of()`, which is implemented on both kinds of IP networks. Better to scope the type restriction to `_BaseNetwork`?

aerleon/lib/nacaddr.py
```diff
@@ -83,13 +98,13 @@ def __init__(self, ip_string, comment='', token='', strict=True):
...
 
-    def subnet_of(self, other):
+    def subnet_of(self, other: "IPv4") -> bool:
         """Return True if this network is a subnet of other."""
         if self.version != other.version:
             return False
         return self._is_subnet_of(self, other)
 
-    def supernet_of(self, other):
+    def supernet_of(self, other: Union[IPv4, IPv6]) -> bool:
```
jtwb, line 107: You can call `ip.supernet_of()` while mixing IPv4 and IPv6? I know there is IPv4 embedding in IPv6 but is that actually used here?

aerleon/lib/nacaddr.py
jtwb: Or `Union[IPv4, IPv6]`

aerleon/lib/naming.py
```diff
@@ -357,7 +360,7 @@ def _GetParents(self, query, query_group):
                 recursive_parents.append(bp)
         return recursive_parents
 
-    def GetNetChildren(self, query):
+    def GetNetChildren(self, query: str) -> List[Union[str, Any]]:
```
jtwb, line 363: Can the "Any" be made more specific? Looking at the code, the return list members all come from `child = item.split('#')[0].strip()` which feels pretty specific.

aerleon/lib/nacaddr.py
ankenyr: Done

aerleon/lib/naming.py
```diff
@@ -504,7 +507,7 @@ def GetPortParents(self, query, proto):
             raise UndefinedPortError('%s/%s is not found in any service tokens' % (query, proto))
         return sorted(matches)
 
-    def GetServiceByProto(self, query, proto):
+    def GetServiceByProto(self, query: str, proto: str) -> List[Union[str, Any]]:
```
jtwb, line 510: Another "Any" stemming from `str.split()`

aerleon/lib/naming.py
```diff
@@ -783,7 +786,7 @@ def ParseYaml(self, file_handle, file_name):
 
         self.ParseDefinitionsObject(file_data, file_name)
 
-    def ParseDefinitionsObject(self, file_data, file_name):
+    def ParseDefinitionsObject(self, file_data: Dict[str, Any], file_name: str) -> None:
```
jtwb, line 789: Another "Any" that maybe could be more specific?

aerleon/lib/nacaddr.py
ankenyr: You can but it will always return false. self.version is either 4 or 6.

ankenyr: Actually something I am unsure about. `version` is defined in _BaseV4 and _BaseV6 which is eventually what `nacaddr.IPv4` and `nacaddr.IPv6` inherit from. Since this is a method on the nacaddr objects should we use those or should we always go as far down the stack and the method would allow us? I would go with the former for methods in a class and the later for methods outside of classes.

aerleon/lib/naming.py
ankenyr: You are right, strangely the tests never return anything but a string.

aerleon/lib/naming.py
ankenyr: Done

aerleon/lib/naming.py
ankenyr: Done

aerleon/lib/nacaddr.py
ankenyr: Ahh good catch, I didn't look very closely it seems.

aerleon/lib/nacaddr.py
jtwb: I think the most important thing is to be communicative. Use the type signature to express what types this code is designed to operate on. Is nacaddr actually designed to be compatible with other subclasses of _BaseNetwork? If it is, it would be good to select types from the `ipaddress` modules. But if you want nacaddr.IPv[4,6] to mainly interoperate with themselves, use those as the acceptable types.

aerleon/lib/nacaddr.py
```diff
@@ -408,12 +444,14 @@ def CollapseAddrList(addresses, complement_addresses=None):
...
-def SortAddrList(addresses):
+def SortAddrList(addresses: List[Union[Any, IPv6, IPv4]]) -> List[Union[Any, IPv6, IPv4]]:
     """Return a sorted list of nacaddr objects."""
     return sorted(addresses, key=ipaddress.get_mixed_type_key)
 
 
-def RemoveAddressFromList(superset, exclude):
+def RemoveAddressFromList(
+    superset: List[Union[IPv4, IPv6]], exclude: Union[IPv4, IPv6]
+) -> List[Union[Any, IPv6, IPv4]]:
```
jtwb, line 454: This Any and the one on line 447 seem suspect also.

jtwb: Removing them (if it is correct to do so) will also remove the import for typing.Any

aerleon/lib/nacaddr.py
```diff
@@ -308,7 +335,13 @@ def _SafeToMerge(address, merge_target, check_addresses):
     return True
 
 
-def _CollapseAddrListInternal(addresses, complements_by_network):
+def _CollapseAddrListInternal(
+    addresses: List[Union[IPv4, IPv6]],
+    complements_by_network: Union[
+        DefaultDict[ipaddress.IPv4Address, List[IPv4]],
+        DefaultDict[ipaddress.IPv6Address, List[IPv6]],
```
jtwb, line 342: Two more DefaultDicts that can just be Dict

aerleon/lib/nacaddr.py
```diff
@@ -286,7 +306,14 @@ def CollapseAddrListPreserveTokens(addresses):
     return [i for sublist in dedup_array for i in sublist]
 
 
-def _SafeToMerge(address, merge_target, check_addresses):
+def _SafeToMerge(
+    address: Union[IPv4, IPv6],
+    merge_target: Union[IPv4, IPv6],
+    check_addresses: Union[
+        TypedDict[ipaddress.IPv4Address, List[IPv4]],
+        TypedDict[ipaddress.IPv6Address, List[IPv6]],
```
jtwb, line 314: TypedDict is not correct here, should just be Dict

jtwb (approved):

PR #252, opened 2023-02-18
Monkeytype the Cisco family.

ankenyr:
This change adds typing to `cisco`, `ciscoasa`, `cisconx`, `ciscoxr`, and `arista`.

### DIFF
--- a/aerleon/lib/cisco.py
+++ b/aerleon/lib/cisco.py
@@ -694,11 +694,11 @@ class Term(aclgenerator.Term):
         """Returns a formatted port string for the range.
 
         Args:
-          port: str list or none, the port range.
-          proto: str representing proto (tcp, udp, etc).
+          port: The port range represented as a tuple..
+          proto: String or int representing a protocol.
 
         Returns:
-          A string suitable for the ACL.
+          A string used to filter a single or range of ports..
         """
         if not port:
             return ''
@@ -743,7 +743,7 @@ class Term(aclgenerator.Term):
         dport: str,
         icmp_type: Union[int, str],
         icmp_code: Union[int, str],
-        option: List[Union[str, Any]],
+        option: List[str],
     ) -> List[str]:
         """Take the various compenents and turn them into a cisco acl line.
 
@@ -756,7 +756,7 @@ class Term(aclgenerator.Term):
           dport: str, the destination port
           icmp_type: icmp-type numeric specification (if any)
           icmp_code: icmp-code numeric specification (if any)
-          option: list or none, optional, eg. 'logging' tokens.
+          option: list of options, eg. 'logging' tokens.
 
         Returns:
           string of the cisco acl line, suitable for printing.
--- a/aerleon/lib/cisco.py
+++ b/aerleon/lib/cisco.py
@@ -17,11 +17,14 @@
 """Cisco generator."""
 
 import ipaddress
-from typing import Union, cast
+from typing import Any, Dict, List, Optional, Set, Tuple, Union, cast
 
 from absl import logging
 
 from aerleon.lib import aclgenerator, addressbook, nacaddr, summarizer
+from aerleon.lib.nacaddr import IPv4, IPv6
+from aerleon.lib.policy import Policy, Term
+from aerleon.lib.summarizer import DSMNet
 
 _ACTION_TABLE = {
     'accept': 'permit',
@@ -62,7 +65,9 @@ class ExtendedACLTermError(Error):
 class TermStandard:
     """A single standard ACL Term."""
 
-    def __init__(self, term, filter_name, platform='cisco', verbose=True):
+    def __init__(
+        self, term: Term, filter_name: str, platform: str = 'cisco', verbose: bool = True
+    ) -> None:
         self.term = term
         self.filter_name = filter_name
         self.platform = platform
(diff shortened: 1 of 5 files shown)


aerleon/lib/cisco.py
```diff
@@ -685,7 +690,7 @@ def _GetIpString(self, addr):
             return 'host %s' % (addr.network_address)
         return addr
 
-    def _FormatPort(self, port, proto):
+    def _FormatPort(self, port: Union[Tuple[()], Tuple[int, int]], proto: Union[int, str]) -> str:
```
jtwb, line 693: What is `Tuple[()]`? Is this a real type we want as input? Shouldn't it be None? Also, the docstring says that port can be "str list" but the type expression is `Tuple[int, int]` - they cannot both be right.

aerleon/lib/cisco.py
```diff
@@ -727,8 +734,17 @@ def _FixOptions(self, proto, option):
...
+        self,
+        action: str,
+        proto: Union[int, str],
+        saddr: str,
+        sport: str,
+        daddr: str,
+        dport: str,
+        icmp_type: Union[int, str],
+        icmp_code: Union[int, str],
+        option: List[Union[str, Any]],
```
jtwb, line 746: Can we remove the Any from `option`? Also the docstring claims `option` can be None but it clearly would crash.

aerleon/lib/cisco.py
jtwb: It should be `Optional[Tuple[int, int]]` just looking at the code here and the code in `PortMap.GetProtocol`.

jtwb: The docstring is definitely incorrect.

aerleon/lib/cisco.py
```diff
@@ -805,7 +821,7 @@ class ObjectGroupTerm(Term):
     in the acl.
     """
 
-    def _FormatPort(self, port, proto):
+    def _FormatPort(self, port: Union[Tuple[()], Tuple[int, int]], proto: str) -> str:
```
jtwb, line 824: Same as above, should be `Optional[Tuple[int, int]]`

aerleon/lib/ciscoasa.py
```diff
@@ -193,8 +197,17 @@ def __str__(self):
...
-        self, filter_name, action, proto, saddr, sport, daddr, dport, icmp_type, option
-    ):
+        self,
+        filter_name: str,
+        action: str,
+        proto: str,
+        saddr: Union[str, IPv4, IPv6],
+        sport: Union[Tuple[()], Tuple[int, int]],
+        daddr: Union[str, IPv4, IPv6],
+        dport: Union[Tuple[()], Tuple[int, int]],
```
jtwb, line 207: sport and dport should be `Optional[Tuple[int, int]]`

aerleon/lib/ciscoxr.py
```diff
@@ -58,7 +61,7 @@ def _BuildTokens(self):
 
         return supported_tokens, supported_sub_tokens
 
-    def _GetObjectGroupTerm(self, term, verbose=True):
+    def _GetObjectGroupTerm(self, term: Term, verbose: bool = True) -> "CiscoXRObjectGroupTerm":
```
jtwb, line 64: Any idea why CiscoXRObjectGroupTerm appears in quotes? Not a problem but it doesn't seem to match the rest of the PR.

aerleon/lib/cisco.py
ankenyr: Yea the docstring is terrible looking at it. I can fix that up in this PR.
`Tuple[()]` is meant to be an empty tuple. 
Some select output from printing `port` and `type(port)`
```
E   + <class 'tuple'>
E   + ()
E   + <class 'tuple'>
E   + (1024, 65535)
E   + <class 'tuple'>
E   + ()
E   + <class 'tuple'>
E   + (1024, 65535)
E   + <class 'tuple'>
E   + ()
E   + <class 'tuple'>
E   + (1024, 65535)
E   + <class 'tuple'>
E   + ()
E   + <class 'tuple'>
```

I don't think it would be valid to do optional since `Optional` would be equal to a union of the type and None.
```
>>> tuple() == None
False
```
And those are not equal.

aerleon/lib/ciscoxr.py
jtwb: I think this just means the type has not yet been defined in this file.

aerleon/lib/cisco.py
jtwb: Ok, that makes sense. I'll resolve the other suggestions to use Optional. I would suggest swapping the order from `Tuple[()], Tuple[int, int]` to `Tuple[int, int], Tuple[()]` for clarity.

aerleon/lib/cisco.py
ankenyr: Done

aerleon/lib/cisco.py
ankenyr: Same response as above. Lets leave the conversation there and we can resolve this based on the one above.

aerleon/lib/ciscoxr.py
ankenyr: Ahh monkeytype has been doing that in some cases and not others. I try to fix it where I see it. Thanks for catching it.

aerleon/lib/cisco.py
jtwb: It would be nice if there was a type hint for "falsy" things!

jtwb (approved):

PR #254, opened 2023-02-20
Monkeytype Juniper family.

ankenyr:


aerleon/lib/juniper.py
```diff
@@ -722,7 +727,14 @@ def CheckTerminatingAction(self):
...
 
-    def _MinimizePrefixes(self, include, exclude):
+    def _MinimizePrefixes(
+        self, include: List[Union[IPv4, Any, IPv6]], exclude: List[Union[IPv4, Any]]
+    ) -> Union[
+        Tuple[List[IPv6], List[Any]],
+        Tuple[List[IPv4], List[IPv4]],
+        Tuple[List[Any], List[Any]],
+        Tuple[List[IPv4], List[Any]],
+    ]:
```
jtwb, line 737: This one needs some love.

ankenyr: Done, while tests only include IPv4 and IPv6 I believe this isn't necessarily true and may fail once we turn on mypy or some other static analysis tool.

aerleon/lib/juniper.py
```diff
@@ -753,10 +760,11 @@ def _MinimizePrefixes(self, include, exclude):
                 if exclude_prefix.subnet_of(include_prefix):
                     exclude_result.append(exclude_prefix)
                     break
-
         return include_result, exclude_result
 
-    def _Comment(self, addr, exclude=False, line_length=132):
+    def _Comment(
+        self, addr: Union[IPv6, IPv4, DSMNet], exclude: bool = False, line_length: int = 132
+    ) -> List[Union[Any, str]]:
```
jtwb, line 767: 1. Looking at the code, I think the return type is `List[str]`.
2. The docstring for argument `addr` is out of date.

aerleon/lib/juniper.py
```diff
@@ -848,7 +856,7 @@ def _Comment(self, addr, exclude=False, line_length=132):
             logging.warning('Ignoring non IPv4 or IPv6 address: %s', addr)
         return rval
 
-    def _Group(self, group, lc=True):
+    def _Group(self, group: List[Union[int, Tuple[int, int], str]], lc: bool = True) -> str:
```
jtwb, line 859: Not necessarily in scope, but just spotted this duplicated expression on line 883:
```
if isinstance(el, str) or isinstance(el, str):
```

aerleon/lib/junipermsmpc.py
```diff
@@ -416,7 +419,7 @@ def _BuildPort(self, ports):
                 port_list.append('%s-%s' % (str(p[0]), str(p[1])))
         return port_list
 
-    def _GenerateApplications(self, filter_name):
+    def _GenerateApplications(self, filter_name: str) -> List[Union[Any, str]]:
```
jtwb, line 422: The return type is `List[str]`

aerleon/lib/junipermsmpc.py
```diff
@@ -399,7 +402,7 @@ def _BuildTokens(self):
         )
         return supported_tokens, supported_sub_tokens
 
-    def _BuildPort(self, ports):
+    def _BuildPort(self, ports: List[Union[Any, List[int], Tuple[int, int]]]) -> List[str]:
```
jtwb, line 405: Monkeytype seems to have noticed this function is called with lists instead of tuples sometimes, and sure enough on line 616 (pre-changes), `_TranslatePolicy` constructs a list instead of a tuple to represent a port range. I'd suggest clamping the type to `List[Tuple[int, int]]` and then cleaning up line 616 to use a tuple.

aerleon/lib/junipersrx.py
```diff
@@ -613,7 +615,7 @@ def _TranslatePolicy(self, pol, exp_info):
 
             self.srx_policies.append((header, new_terms, filter_options))
 
-    def _FixLargePolices(self, terms, address_family):
+    def _FixLargePolices(self, terms: List[Any], address_family: str) -> None:
```
jtwb, line 618: "terms" is `List[policy.Term]`.

aerleon/lib/junipersrx.py
```diff
@@ -675,7 +677,7 @@ def Chunks(addresses):
             del terms[:]
             terms.extend(expanded_terms)
 
-    def _BuildPort(self, ports):
+    def _BuildPort(self, ports: List[Union[Any, Tuple[int, int]]]) -> List[Union[Any, str]]:
```
jtwb, line 680: Both "Any" hints can be removed. The type for "ports" is `List[Tuple[int, int]]` and the return type is `List[str]`

ankenyr: done

aerleon/lib/junipersrx.py
ankenyr: done

aerleon/lib/junipermsmpc.py
ankenyr: done

aerleon/lib/junipermsmpc.py
ankenyr: done

aerleon/lib/juniper.py
ankenyr: done

aerleon/lib/juniper.py
ankenyr: done

PR #256, opened 2023-02-21
Hoist unseed check to intermediate representation

### LINKED ISSUE #138: Simplify Naming
1. A single call to _CheckUnseen should check both unseen networks and services.
2. Consolidate duplicate "unseen name" logic that appears in _ParseLine and ParseDefinitionsObject.

btriller:
Also raise NamingSyntaxError for invalid port syntax for multiline services.

fixes #138

### DIFF
--- a/aerleon/lib/naming.py
+++ b/aerleon/lib/naming.py
@@ -47,13 +47,15 @@ DNS = 53/tcp
 
 
 import re
+import sys
 from pathlib import Path
 from typing import Any, Dict, List, Tuple, Union
-import sys
+
 if sys.version_info > (3, 11):
     from typing import Self
 else:
     from typing import TypeVar
+
     Self = TypeVar("Self", bound="_ItemUnit")
 
 import yaml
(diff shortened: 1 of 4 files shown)


jtwb:
Thanks for taking this on. What I had hoped we could do with the duplicate code is to not change `_ItemUnit`, but instead move the duplicate code out of `_ParseLine`, `_ParseYamlServices`, and `_ParseYamlNetworks` into a common error checking function that accepts `_ItemUnit` as input.

Something like this:
```
class Naming:
# ...
    def AddUnit(item: _ItemUnit, def_type: str):
            if not self.TOKEN_RE.match(item.name):
                # Vary message by def type
                logging.info(
                    f'\nService name does not match recommended criteria: {item.name}\nOnly A-Z, a-z, 0-9, -, and _ allowed'
                )
            if item.name in self.services: # Vary target and message by def_type
                raise NamespaceCollisionError(
                    f'\nMultiple definitions found for service: {item.name}'
                )
            self.services[symbol] = item # Vary by def_type (or just have a different function for Symbol vs Network)
            if item.name in self.unseen_services:
                self.unseen_services.pop(item.name)
```

aerleon/lib/naming.py
```diff
@@ -663,111 +706,65 @@ def ParseNetworkList(self, data: List[str]) -> None:
           data: array of text lines containing net definitions.
 
         """
-        for line in data:
-            self._ParseLine(line, DEF_TYPE_NETWORKS)
-
-    def _ParseLine(self, line: str, definition_type: str) -> None:
```
jtwb, line 669: I'm concerned it might not be safe to remove `Naming._ParseLine` in case network automation tools or scripts are using it to load data one line at a time. I suspect it might be in use because it is easy to loop over your input and call `Naming._ParseLine` each time than to construct a whole file and call `Naming.ParseNetworkList` (even though it has an underscore name).

ankenyr:
hey @btriller did you have time to respond?

PR #257, opened 2023-02-25
Monkeytyping core libaries.

ankenyr:
This PR monkey types a set of our core libraries 
aerleon/lib/aclgenerator.py
 aerleon/lib/plugin_supervisor.py
aerleon/lib/policy.py
aerleon/lib/policy_simple.py
aerleon/lib/recognizers.py
aerleon/lib/summarizer.py
aerleon/lib/yaml.py

Some functions were not hit due to not being covered by tests. A follow-up PR should either create more tests or create typing manually in the future.

### DIFF
--- a/aerleon/lib/policy.py
+++ b/aerleon/lib/policy.py
@@ -2754,8 +2754,6 @@ def ParsePolicy(
     Returns:
       policy object or False (if parse error).
     """
-    print("definitions")
-    print(type(definitions))
     try:
         if definitions:
             globals()['DEFINITIONS'] = definitions
--- a/aerleon/lib/policy.py
+++ b/aerleon/lib/policy.py
@@ -22,12 +22,16 @@ import datetime
 import os
 import pathlib
 import sys
-from typing import TYPE_CHECKING
+from typing import Any, List, Optional, Tuple, Union, TYPE_CHECKING
 
 from absl import logging
 from ply import lex, yacc
 
 from aerleon.lib import nacaddr, naming
+from aerleon.lib.nacaddr import IPv4, IPv6
+from ply.lex import LexToken
+from ply.yacc import YaccProduction
+from unittest.mock import MagicMock
 
 if TYPE_CHECKING:
     from aerleon.lib.policy_builder import PolicyBuilder
@@ -141,7 +145,9 @@ class InvalidNumericProtoValue(Error):
     """Error when protocols are numeric and not between -1 and 255."""
 
 
-def TranslatePorts(ports, protocols, term_name):
+def TranslatePorts(
+    ports: List[str], protocols: List[Union[str]], term_name: str
+) -> List[Tuple[int, int]]:
     """Return all ports of all protocols requested.
 
     Args:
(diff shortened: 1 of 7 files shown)


aerleon/lib/policy.py
```diff
@@ -141,7 +145,9 @@ class InvalidNumericProtoValue(Error):
     """Error when protocols are numeric and not between -1 and 255."""
 
 
-def TranslatePorts(ports, protocols, term_name):
+def TranslatePorts(
+    ports: List[str], protocols: List[Union[str]], term_name: str
```
jtwb, line 149: `Union[str]` is just `str`

aerleon/lib/policy.py
```diff
@@ -2592,7 +2606,7 @@ def p_strings_or_ints(p):
             p[0] = [p[1]]
 
 
-def p_error(p):
+def p_error(p: LexToken):
```
jtwb, line 2609: Good catch by monkeytype here - the PLY docs consistently refer to the argument to `p_error` as "p" even though it is a LexToken and not a production. Would be more correct as "t".

aerleon/lib/policy.py
```diff
@@ -2718,8 +2732,13 @@ def ParseFile(filename, definitions=None, optimize=True, base_dir='', shade_chec
 
 
 def ParsePolicy(
-    data, definitions=None, optimize=True, base_dir='', shade_check=False, filename=''
-):
+    data: str,
+    definitions: Optional[MagicMock] = None,
```
jtwb, line 2736: This is not correct, MagicMock should not be part of the public interface.

aerleon/lib/recognizers.py
```diff
@@ -345,7 +345,7 @@ class TSection(TComposition):
 
     of: "list[typing.Tuple[str | TValue | TUnion, TValue | TComposition]]"
 
-    def Recognize(self, value: dict):
+    def Recognize(self, value: dict) -> typing.Dict[str, typing.List[str]]:
```
jtwb, line 348: We should probably just import Dict, List and Any at this point.

aerleon/lib/yaml.py
```diff
@@ -145,7 +146,9 @@ def ParsePolicy(
     return policy.FromBuilder(PolicyBuilder(policy_dict, definitions, optimize, shade_check))
 
 
-def PreprocessYAMLPolicy(filename, base_dir, policy_dict: PolicyDict):
+def PreprocessYAMLPolicy(
+    filename: str, base_dir: str, policy_dict: PolicyDict
+) -> Optional[Dict[str, List[Dict[str, Union[Dict[str, Dict[str, str]], List[Dict[str, str]]]]]]]:
```
jtwb, line 151: This should just be `Optional[PolicyDict]`

jtwb (approved):
Apart from comments this PR looks very good.

aerleon/lib/policy.py
ankenyr: Done

aerleon/lib/policy.py
ankenyr: Done

aerleon/lib/yaml.py
ankenyr: Done

PR #260, opened 2023-03-10
[Typing] Initial Cloudarmor typing.

ankenyr:


aerleon/lib/cloudarmor.py
```diff
@@ -137,14 +152,10 @@ def ConvertToDict(self, priority_index):
                 }
                 rules.append(rule)
 
-        # TODO(robankeny@): Review this log entry to make it cleaner/more useful.
-        # Right now, it prints the entire term which might be huge
         if len(source_addr_chunks) > 1:
             logging.debug(
-                'Current term [%s] was split into %d sub-terms since '
```
jtwb, line 144: Did you mean to drop the braces?

jtwb (approved):

aerleon/lib/cloudarmor.py
ankenyr: In this case yea, it reads better imo

PR #262, opened 2023-03-10
[Typing] Initial gcp_hf.py typing

ankenyr:


aerleon/lib/gcp_hf.py
```diff
@@ -539,7 +570,7 @@ def _TranslatePolicy(self, pol, exp_info):
             logging.info('Policy %s quota cost: %d', policy[display_name], total_cost)
 
 
-def GetRuleTupleCount(dict_term: Dict[str, Any], api_version):
```
jtwb, line 542: I think this Dict[str] is part of the 3.7 failure

jtwb (approved):

PR #266, opened 2023-03-10
[Typing] Initial openconfig typing

ankenyr:


jtwb (approved):

aerleon/lib/openconfig.py
```diff
@@ -45,10 +52,25 @@ class ExceededAttributeCountError(Error):
 
 
 # Graceful handling of dict heierarchy for OpenConfig JSON.
-def RecursiveDict():
+def RecursiveDict() -> DefaultDict[Any, Any]:
```
jtwb, line 55: Is it possible to just use `-> DefaultDict` ?

PR #267, opened 2023-03-10
[Typing] Packetfilter typing

ankenyr:


aerleon/lib/packetfilter.py
```diff
@@ -282,22 +291,22 @@ def _CheckAddressAf(self, addrs):
...
-        stateful,
-    ):
+        action: str,
+        direction: str,
+        log: List[VarType],
+        interface: Optional[str],
+        af: str,
+        proto: List[str],
+        src_addr: str,
+        src_port: Union[List[Any], str],
```
jtwb, line 301: Can we constrain the Any here? Is this just a list of nacaddr objects?

ankenyr: List of strings actually since these are ports. Nacaddr are for addresses only.

aerleon/lib/packetfilter.py
```diff
@@ -18,11 +18,13 @@
 
 import collections
 import copy
-from typing import cast
+from typing import Any, Dict, List, Optional, Set, Tuple, Union, cast
```
jtwb, line 21: Looks like 'Any' is imported but unused

jtwb (approved):

aerleon/lib/packetfilter.py
ankenyr: Done

PR #268, opened 2023-03-10
[Typing] Initial paloaltofw typing.

ankenyr:


aerleon/lib/paloaltofw.py
```diff
@@ -119,7 +128,12 @@ def __init__(self, from_zone, to_zone, term, service_map):
             self.options.append(x)
 
     @staticmethod
-    def TermToOptions(from_zone, to_zone, term, service_map):
+    def TermToOptions(
+        from_zone: str, to_zone: str, term: Term, service_map: ServiceMap
+    ) -> Union[
+        Tuple[Dict[str, Union[List[str], str]], None],
+        Tuple[Dict[str, Union[List[str], str]], Term],
```
jtwb, line 135: Could this just be `-> Tuple[Dict[str, Union[List[str], str]], Optional[Term]]` ?

ankenyr: well boy sure howdy it could.

jtwb (approved):

PR #269, opened 2023-03-10
[Typing] Initial Pcap typing

ankenyr:


aerleon/lib/pcap.py
```diff
@@ -192,7 +198,7 @@ def __str__(self):
 
         return cond + '\n'
 
-    def _CheckAddressAf(self, addrs):
+    def _CheckAddressAf(self, addrs: List[Union[Any, IPv6, IPv4]]) -> List[Union[str, IPv6, IPv4]]:
```
jtwb, line 201: That 'Any' can be removed, I think.

aerleon/lib/pcap.py
```diff
@@ -206,7 +212,7 @@ def _CheckAddressAf(self, addrs):
         return af_addrs
 
     @staticmethod
-    def JoinConditionals(condition_list, operator):
+    def JoinConditionals(condition_list: List[Union[Any, str]], operator: str) -> str:
```
jtwb, line 215: Same, this Any can be removed

aerleon/lib/pcap.py
```diff
@@ -228,7 +234,9 @@ def JoinConditionals(condition_list, operator):
         res = '(%s)' % (op.join(condition_list))
         return res
 
-    def _GenerateAddrStatement(self, addrs, exclude_addrs):
+    def _GenerateAddrStatement(
+        self, addrs: List[Union[str, IPv6, IPv4]], exclude_addrs: List[Any]
```
jtwb, line 238: `exclude_addrs` should have the same type as `addrs`, right?

aerleon/lib/pcap.py
```diff
@@ -266,7 +276,7 @@ def _GeneratePortStatement(self, ports, direction):
                 conditions.append('%s portrange %s-%s' % (direction, port_tuple[0], port_tuple[1]))
         return Term.JoinConditionals(conditions, 'or')
 
-    def _GenerateTcpOptions(self, options):
+    def _GenerateTcpOptions(self, options: List[Union[Any, str]]) -> str:
```
jtwb, line 279: This is just `List[str]` since it only is ever called with Term.options.

aerleon/lib/pcap.py
ankenyr: Done

aerleon/lib/pcap.py
ankenyr: Yup, think these were missed during the late night coding.

aerleon/lib/pcap.py
ankenyr: Done

jtwb (approved):

PR #271, opened 2023-03-10
[Typing] Initial Windows typing

ankenyr:


aerleon/lib/windows.py
```diff
@@ -191,7 +195,7 @@ def _HandlePorts(self, src_ports, dst_ports):
         """
         return None, None
 
-    def _HandlePreRule(self, ret_str):
+    def _HandlePreRule(self, ret_str: List[Union[str, Any]]) -> None:
```
jtwb, line 198: This should just be `List[str]`. We should also add the same type signature to _HandlePreRule in windows_ipsec.py .

aerleon/lib/windows_advfirewall.py
```diff
@@ -75,12 +79,21 @@ def _HandleIcmpTypes(self, icmp_types, protocols):
 
         return (types, protocols)
 
-    def _HandlePorts(self, src_ports, dst_ports):
+    def _HandlePorts(
+        self, src_ports: List[Any], dst_ports: List[Union[Tuple[int, int], Any]]
```
jtwb, line 83: Anywhere Monkeytype generates `Union[..., Any]` we should try to get rid of the "Any" and figure out what the type requirement really is.

aerleon/lib/windows.py
ankenyr: Done

jtwb (approved):

PR #274, opened 2023-03-21
[Typing] Typing functions missed by monkeytype.

ankenyr:
This PR creates a ton of type hinting across the code base. Some things to note for future improvements.

- `_Group` is a function that is common across multiple generators. There are some slight differences but mostly it is copy/paste. We should bring this out of those.
- `ports.py` was mostly skipped. PPP looks to be mostly unused except in one function in `naming.py` which is unused. In general I think we should be creating a real port object similar to our discussions of a better nacaddr object and fixing the usage of strings being passed into places a nacaddr object should be passed in.
- To differentiate `policy.Term` from `aclgenerator.Term` I have changed many generators to import policy rather than the objects within.
- `nftables.py` has some missing functions. The test coverage here is low (~70%) so I am unconfident in what is being passed into every function. I could guess but I would rather add more tests and ensure this.
- Fixed a double `isinstance(el, str) in `juniper.py` that was left over from the py2 to py3 conversion.
- Removed a few comments that are no longer relevant.
- Removed unused imports
- In a few instances removed double quotes around type hints.
- Fixed unused import in `aruba.py` by patching `logging` in its parent `aclgenerator` rather than importing it unused in `aruba.py`.

### DIFF
--- a/aerleon/lib/nftables.py
+++ b/aerleon/lib/nftables.py
@@ -768,7 +768,7 @@ class Nftables(aclgenerator.ACLGenerator):
         return netfilter_family, netfilter_hook, netfilter_priority, policy_default_action, verbose
 
     def _ConfigurationDictionary(
-        self, nft_pol: List[Union[policy.Header, str, str, str, int, str, bool, DefaultDict]]
+        self, nft_pol: List[Union[policy.Header, str, int, bool, DefaultDict]]
     ) -> DefaultDict[str, Union[str, DefaultDict]]:
         """NFTables configuration object.
 
--- a/aerleon/lib/nftables.py
+++ b/aerleon/lib/nftables.py
@@ -17,10 +17,11 @@
 
 import collections
 import copy
+from typing import DefaultDict, Dict, List, Set, Tuple, Union
 
 from absl import logging
 
-from aerleon.lib import aclgenerator, nacaddr
+from aerleon.lib import aclgenerator, nacaddr, policy
 
 # NFTables and Aerleon have conflicting definitions of 'address family'
 # In Aerleon:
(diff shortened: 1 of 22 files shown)


aerleon/lib/summarizer.py
```diff
@@ -43,7 +43,8 @@ def __init__(self, address: int, netmask: int, text: str = '') -> None:
         self.netmask = netmask
         self.text = text
 
-    def __hash__(self):
+    def __hash__(self) -> int:
+        print(type(hash(str(self.address) + str(self.netmask))))
```
jtwb, line 47:
```suggestion
```

aerleon/lib/summarizer.py
```diff
@@ -52,19 +53,20 @@ def __eq__(self, other: DSMNet) -> bool:
         except AttributeError:
             return NotImplemented
 
-    def __ne__(self, other):
+    def __ne__(self, other: DSMNet) -> bool:
+        print(type(other))
```
jtwb, line 57:
```suggestion
```

aerleon/lib/nftables.py
```diff
@@ -756,7 +767,9 @@ def _ProcessHeader(self, header_options):
             header_options.remove('noverbose')
         return netfilter_family, netfilter_hook, netfilter_priority, policy_default_action, verbose
 
-    def _ConfigurationDictionary(self, nft_pol):
+    def _ConfigurationDictionary(
+        self, nft_pol: List[Union[policy.Header, str, str, str, int, str, bool, DefaultDict]]
```
jtwb, line 771: This doesn't look right, this is a Union with duplicate entries:
```
Union[policy.Header, str, str, str, int, str, bool, DefaultDict]
```

ankenyr: Ahh I should have added this one to the notes. I just changed it to be correct for now. This should actually be a tuple eventually as that would make more sense to pass in rather than a list of a bunch of different types.

jtwb (approved):

PR #275, opened 2023-03-22
Removing dead code from nacaddr.py.

ankenyr:
Back in when this project was first created Python had no ipaddress library. IIRC Peter Moody created it and open sourced it which is why you will find it copyrighted by Google in https://github.com/python/cpython/blob/main/Lib/ipaddress.py#L1276

You can see there are some slight differences between the one in the "third party" folder and the open source version which is what these deleted lined helped to bridge.
https://github.com/google/capirca/commit/4f694aa1af94352e6f89a87bb85ac201488b4967

I also deleted a function ` _is_subnet_of` that was necessary while Google was stuck in pre 3.7 but is included at 3.7. We only support 3.7+ so it is safe to remove ` _is_subnet_of`.

### DIFF
--- a/aerleon/lib/nacaddr.py
+++ b/aerleon/lib/nacaddr.py
@@ -21,9 +21,31 @@ from __future__ import annotations
 import collections
 import ipaddress
 import itertools
+import warnings
 from typing import Union, cast
 
 import aerleon.utils.iputils as iputils
+
+# Names kept only for backwards compatibility with Capirca / nacaddr v1. Calling
+# one emits a DeprecationWarning pointing at its replacement.
+# TODO(aerleon): remove the deprecated names (IPv4.Supernet, IPv6.Supernet,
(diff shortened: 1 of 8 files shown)


aerleon/lib/nacaddr.py
```diff
@@ -516,12 +492,5 @@ def AddressListExclude(
         return sorted(set(ret_array + superset))
 
 
-ExcludeAddrs = AddressListExclude
```
jtwb, line 519: Seems like third party generators could still be using ExcludeAddrs and Supernet, especially if they looked at code in the built-in generators using them. A deprecation notice seems warranted.

jtwb: I should say, a deprecation notice would be warranted and we should not remove ExcludeAddrs and Supernet right away.

jtwb (changes requested):
See deprecation notice comment.

PR #279, opened 2023-04-25
Adding copyright to addressbook.py

ankenyr:
Opps


jtwb:
New files added to the Aerleon repo with no copyright headers are OK as long as they do not contain any content that was present in a file with a copyright header (in which case we would need to include a copy of that copyright header). In the case of addressbook.py, every line of code was authored by Aerleon contributors, so this file should not have any header.

PR #281, opened 2023-05-03
[WIP] Adding initial code to support hostnames.

ankenyr:
This adds functionality to `policy.py`, `addressbook.py`, and `policy_builder.py` to support hostnames.

Usage of these new keywords are similar to address keywords. `source-address` and `destination-address` will accept a reference to a network definition. That network definition will need to have a hostname entry.

Note that the implementation of these features may vary across different platforms, so their usage within each platform depends on the specific implementation.


ankenyr:
Closing this in favor of a cleaner PR

Issue #282, opened 2023-05-04
Cisco ASA syntax is not correct for /32 destinations


mf84:
This was tested on version 1.2.2. If we start with the example ACLs outlined in the documentation (https://aerleon.readthedocs.io/en/latest/getting_started/) generating ASA ACLs works correctly.

However if I modify the policy in a way that it has hostroutes (/32) as destination address, aerleon will generate syntax that is not accepted by ASA.

```
filters:
- header:
    comment: Example inbound
    targets:
      cisco: inbound extended
      ciscoasa: test
  terms:
    - name: accept-web-servers
      comment: Accept connections to our web servers
      destination-address: WEB_SERVERS
      destination-port: WEB
      protocol: tcp
      action: accept
    - name: test
      comment: both src+dst are /32
      source-address: WEB_SERVERS
      destination-address: MAIL_SERVERS
      action: accept
    - name: default-deny
      comment: Deny anything else.
      action: deny
```

The resulting file example.pol.asa looks like this:

```
clear configure access-list test
access-list test remark $Id:$
access-list test remark $Date:$
access-list test remark $Revision:$
access-list test remark Example inbound


access-list test remark accept-web-servers
access-list test remark Accept connections to our web servers
access-list test extended permit tcp any host 10.0.0.1 eq www
access-list test extended permit tcp any host 10.0.0.1 eq https
access-list test extended permit tcp any host 10.0.0.2 eq www
access-list test extended permit tcp any host 10.0.0.2 eq https


access-list test remark test
access-list test remark both src+dst are /32
access-list test extended permit ip host 10.0.0.1 10.0.0.3/32
access-list test extended permit ip host 10.0.0.1 10.0.0.4/32
access-list test extended permit ip host 10.0.0.2 10.0.0.3/32
access-list test extended permit ip host 10.0.0.2 10.0.0.4/32


access-list test remark default-deny
access-list test remark Deny anything else.
access-list test extended deny ip any any
```

Note that instead of having `access-list test extended permit ip host 10.0.0.1 host 10.0.0.3` as line, the created code reads
`access-list test extended permit ip host 10.0.0.1 10.0.0.3/32`. This is not valid ASA syntax.

For some reason /32s are done correctly when used as source, but not when used as destination.

ankenyr:
Thank you for the report! This likely is a problem with `_GetIpString` and what is being fed into it.

jtwb:
[...]

PR #283, opened 2023-05-07
PR282 Fixes destination hosts being outputted incorrectly.

ankenyr:
The term to str function currently does if/elif rather than sequential if statements. Regression test added to prevent this from happening during any cleanup we may perform in the future.


ankenyr:
This should fix https://github.com/aerleon/aerleon/pull/283

jtwb (approved):

jtwb:
Looks like a good fix for #282. As a follow up, can we have a test for the DSMNet cases? Partially for my own enlightenment since I'm not totally familiar with the desired behavior there.

PR #290, opened 2023-05-18
Add devcontainer configuration

jtwb:
Aerleon just relies on Python and PyPoetry, so it doesn't require much setup for developers to get started contributing to the application. But there has been some interest in having a Docker-based development workflow.

This PR integrates with VSCode's "Dev Containers" feature. When opening this project in VSCode, the IDE will prompt users to "Reopen this project in a container", which will start a Docker container with Python and PyPoetry installed.


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,28 @@
...
+	// "forwardPorts": [],
+
+	// Use 'postCreateCommand' to run commands after the container is created.
+	// "postCreateCommand": "pip3 install --user -r requirements.txt",
+
+	// Configure tool-specific properties.
+	// "customizations": {},
+
+	// Uncomment to connect as root instead. More info: https://aka.ms/dev-containers-non-root.
+	// "remoteUser": "root"
```
ankenyr, line 27: Did we want to keep these commented out fields or clean this up?

.gitignore
```diff
@@ -113,7 +113,6 @@ celerybeat.pid
 *.sage.py
 
 # Environments
-.env
```
ankenyr, line 116: was this intended?

extensions.json
```diff
@@ -0,0 +1,13 @@
...
+
+	// List of extensions which should be recommended for users of this workspace.
+	"recommendations": [
+		"ms-vscode-remote.remote-containers",
+	],
+	// List of extensions recommended by VS Code that should not be recommended for users of this workspace.
+	"unwantedRecommendations": [
+		
+	]
+}
```
ankenyr, line 13: nit: new line here.

ankenyr (commented):
Made a few comments, thanks Jason.

extensions.json
jtwb: Fixed

.gitignore
jtwb: Yes, the devcontainer extension uses a .env file for configuration. This file would need to be checked in to the repo. Our setup doesn't use that file right now, but we should alter the gitignore so there are no surprises for future users.

https://stackoverflow.com/a/58227846

.devcontainer/devcontainer.json
jtwb: Keep them. This is how VSCode's tooling generates these files. I don't think JSON was a good choice on VSCode's part but it is what it is.

ankenyr (approved):

PR #291, opened 2023-05-25
Adding FQDN to the core libraries.

ankenyr:
This change adds FQDN to the core libraries.

FQDN's can now be added to a network definition by adding a `fqdn`  entry into the values similar to `address`. The following is an example

```
  FOOBAR:
    values:
      - 9OCLOCK
      - fqdn: foo.bar
        comment: "foo's fqdn"
```

The FQDN object takes after the nacaddr object in that it has a `token`, `parent_token`, `text` for comments, and the field that contains the value.

Follow-up PR will implement usage of this into generators themselves.


aerleon/lib/fqdn.py
```diff
@@ -0,0 +1,19 @@
+class FQDN:
```
jtwb, line 1: It looks like FQDN could just be a Python dataclass. The default __eq__ method generated by the dataclass decorator will be identical to your implementation here, so you can remove it.

```python
from dataclasses import dataclass

@dataclass
class FQDN:
    """Docstring pls"""
    fqdn: str
    text: float
    token: token
    parent_token: str
    
    def __init__(self, fqdn: str, token: str, comment: str = ''):
        self.fqdn = fqdn
        self.text = comment
        self.token = token
        self.parent_token = token
```


https://docs.python.org/3/library/dataclasses.html

aerleon/lib/naming.py
```diff
@@ -216,10 +217,13 @@ def __init__(
         self.current_symbol = None
         self.services = {}
         self.networks = {}
+        self.fqdn = {}
         self.unseen_services = {}
         self.unseen_networks = {}
         self.port_re = re.compile(r'(^\d+-\d+|^\d+)\/\w+$|^[\w\d-]+$', re.IGNORECASE | re.DOTALL)
         self.token_re = re.compile(r'(^[-_A-Z0-9]+$)', re.IGNORECASE)
+        # https://regexr.com/3g5j0
```
jtwb, line 225: The test cases on regexer are nice. Time permitting, might be nice to move them into small Python tests.

aerleon/lib/policy_builder.py
```diff
@@ -923,4 +931,4 @@ def _NormalizeValues(cls, context: RecognizerContext, repr):
             # by the user.
             repr = (repr['name'], repr.get('policy', ''))
 
-        return repr
+        return repr
```
jtwb, line 934: Nitpick: final newline is missing. Also the indentation on the data changes to _Builtin and _BuildTerm do not match the surrounding lines.

aerleon/lib/naming.py
```diff
@@ -538,6 +542,29 @@ def GetServiceByProto(self, query: str, proto: str) -> List[str]:
...
+        for i in self.fqdn[token].items:
+            comment = ''
+            if i.find('#') > -1:
+                (name, comment) = i.split('#', 1)
+            else:
+                name = i
+
+            name = name.strip()
+            # Check if hostname is actually a subtoken
+            if self.token_re.match(name):
```
jtwb, line 560: Thinking out loud here: is there any possible overlap between token_re and fqdn_re?
```
(^[-_A-Z0-9]+$)
```
vs
```
^(?!:\/\/)(?=.{1,255}$)((.{1,63}\.){1,127}(?![0-9]*$)[a-z0-9-]+\.?)$
```

jtwb: Seems the answer is no: periods cannot appear in token names but are required for FQDN names.

aerleon/lib/naming.py
```diff
@@ -538,6 +542,29 @@ def GetServiceByProto(self, query: str, proto: str) -> List[str]:
...
+                returnlist.extend(self.GetFQDN(name))
+            elif self.fqdn_re.match(name):
+                returnlist.append(FQDN(name, token, comment))
+            else:
+                logging.debug(f"{name} did not match FQDN")
```
jtwb, line 565: In GetNet, we have the IP constructor raise an error if the input is not a valid IP address. Can we move the FQDN validation logic into the FQDN constructor to mirror how we use the IP constructor? Then, `fqdn_re` can be encapsulated as it is a concern of the FQDN constructor and not the naming library.

aerleon/lib/naming.py
```diff
@@ -877,11 +909,26 @@ def _ParseYamlNetworks(self, file_data: Dict[str, Any], file_name: str) -> None:
...
+                        fqdn_unit.items.append(f'{value} # {comment}')
+                    elif ip:
+                        addr_unit.items.append(f'{value} # {comment}')
+                    else:
+                        fqdn_unit.items.append(f'{value} # {comment}')
+                    
 
-                if network_ref and network_ref not in self.networks:
+                if network_ref and (
+                    network_ref not in self.networks or network_ref not in self.fqdn
```
jtwb, line 930: This logic change isn't strictly necessary since line 874 always creates an entry in self.networks for a symbol even if there is only FQDN data for that symbol.

aerleon/lib/naming.py
```diff
@@ -877,11 +909,26 @@ def _ParseYamlNetworks(self, file_data: Dict[str, Any], file_name: str) -> None:
                     continue
 
                 if comment is None:
-                    unit.items.append(value)
+                    if network_ref:
+                        addr_unit.items.append(value)
```
jtwb, line 913: I spent some time looking at how Naming.GetNet will behave when a network refers to a network with FQDN data, or when a network contains mixed data. Mostly the behavior looks reasonable, but this is complex enough that some extra unit tests are needed.

Some cases worth testing:
* A network which references only FQDN-only networks.
* A network which references networks with a mix of FQDN-only, mixed, and IP-only networks.
* A network with only FQDN data.
* A network with mixed data (FQDN and IP items).

I also spotted a corner case that might surprise users: in the following example, GetNet('EXAMPLE_NETWORK') returns an empty list, which the generators will interpret to mean that `source-address` was not present at all in the term. In this example, source-address is not included at all in the generated config. Would it be better to crash in this case? Or generate a warning?

```
terms:
- name: example-term
  source-address: EXAMPLE_NETWORK
...
networks:
  EXAMPLE_NETWORK:
    fqdn: example.com
```

aerleon/lib/naming.py
```diff
@@ -840,10 +867,12 @@ def _ParseYamlNetworks(self, file_data: Dict[str, Any], file_name: str) -> None:
                     f'\nMultiple definitions found for network: {symbol}'
                 )
 
-            unit = _ItemUnit(symbol)
+            addr_unit = _ItemUnit(symbol)
+            fqdn_unit = _ItemUnit(symbol)
 
             # TODO(jb) This operation should be performed on the IR so we can hoist it from _ParseLine
```
jtwb, line 873: Might be best to delete this comment since the addition of fqdn means that _ParseYamlNetworks no longer mirrors _ParseLine.

jtwb:
Can we update the docs to mention FQDN in the follow-up PR?

aerleon/lib/fqdn.py
ankenyr: Done

aerleon/lib/naming.py
ankenyr: Done

aerleon/lib/naming.py
ankenyr: Done

aerleon/lib/naming.py
ankenyr: Done

aerleon/lib/naming.py
ankenyr: They should be the same now.

aerleon/lib/naming.py
```diff
@@ -538,6 +540,34 @@ def GetServiceByProto(self, query: str, proto: str) -> List[str]:
...
+                name = i
+
+            name = name.strip()
+            if self.token_re.match(name):
+                returnlist.extend(self.GetFQDN(name))
+            else:
+                try:
+                    fqdn = FQDN(name, token, comment)
+                    fqdn.text = comment.lstrip()
+                    fqdn.token = token
```
jtwb, line 563: Except for calling lstrip on comment, I think this is already handled in `FQDN.__init__` lines 15-18

jtwb (approved):

aerleon/lib/naming.py
ankenyr: Should have all the cases covered for FQDN and other changes.

aerleon/lib/naming.py
ankenyr: It was but I removed it because it will then be in line with the other function. This should likely be fixed in nacaddr and fqdn in a followup.

aerleon/lib/naming.py
jtwb: Yeah, and I like that we are going to crash now for GetNet / GetFQDN calls that produce no results.

jtwb:
Changes look good, PR is in good shape

PR #295, opened 2023-05-28
Help text and docs update

jtwb:
This PR clarifies some of aclgen's --helpful content and expands the "CLI Reference" page in the docs.

The proposed markdown content and updated help output can be viewed here: https://github.com/aerleon/aerleon/blob/schema2/docs/reference/cli.md

### DIFF
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -40,7 +40,8 @@ exp_info | ✔ |  |
   --config_file: A YAML file with configuration options;
     repeat this option to specify a list of values
 
-  --[no]debug: Display detailed messages. Default: 'false'
+  --[no]debug: Display detailed messages.
+    Default: 'false'
 
   --definitions_directory: Directory containing network and service definition files.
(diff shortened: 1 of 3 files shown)


docs/reference/cli.md
```diff
@@ -1,40 +1,184 @@
...
+  --config_file: A YAML file with configuration options;
     repeat this option to specify a list of values
-  --[no]debug: Debug messages
-    (default: 'false')
-  --definitions_directory: Directory where the definitions can be found.
-    (default: './def')
-  --exp_info: Print a info message when a term is set to expire in that many weeks.
-    (default: '2')
+
+  --[no]debug: Display detailed messages. Default: 'false'
```
ankenyr, line 43: Some of these have Default on a new line, others don't. Do we want to adopt a style for this with regards to documentation and the help output?

jtwb: Defaults should be on a new line always. Good catch

PR #300, opened 2023-06-04
Update FQDN validation

jtwb:
* Updates the FQDN regex in fqdn.py to disallow URL schemes, e.g. `https://foo.bar` is forbidden but `foo.bar` is allowed.
* Enables the `fqdn:` field in the definition YAML schema.

### DIFF
--- a/schemas/aerleon-definitions.schema.json
+++ b/schemas/aerleon-definitions.schema.json
@@ -34,6 +34,11 @@
       "description": "Specifies an IP address or CIDR IP address range expression.",
       "type": "string"
     },
+    "fqdn": {
+      "description": "Specifies a fully qualified domain name with at least two labels.",
+      "type": "string",
+      "pattern": "^(?!:\/\/)(?=.{1,255}$)((.{1,63}.){1,127}(?![0-9]*$)[a-z0-9-]+.?)$"
+    },
     "port": {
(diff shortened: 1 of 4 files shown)


schemas/aerleon-definitions.schema.json
```diff
@@ -34,6 +34,11 @@
       "description": "Specifies an IP address or CIDR IP address range expression.",
       "type": "string"
     },
+    "fqdn": {
+      "description": "Specifies a fully qualified domain name with two or more labels.",
+      "type": "string",
+      "pattern": "^(?!:\/\/)(?=.{1,255}$)((.{1,63}.){1,127}(?![0-9]*$)[a-z0-9-]+.?)$"
```
ankenyr, line 40: `(.{1,63}\.)`
I assume the escape on the period isn't necessary for this one?

ankenyr: Also same for the other escaped period from where this was copied from.

ankenyr (approved):

schemas/aerleon-definitions.schema.json
jtwb: Fixed, good catch

ankenyr (approved):

Issue #304, opened 2023-06-14
sip alias missing on Arista running 4.26.7M


wahmad-godaddy:
Hello, 

I am facing an issue with the .eacl files generated by Aerleon that are using "sip" as an alias for port 5060. However, on the Arista device, this alias is missing and hence the input is considered invalid. For information, I am running 4.26.7M on DCS-7280CR3-36S-F.

This is the service token that is used for SIP: 

```SIP = 5060/tcp 5060/udp 5060/sctp # SIP```

and this is how the term inside the policy file looks like:-
```
term SIP {
    source-address:: XYZ
    destination-address:: ABC
    protocol:: udp
    source-port:: HIGH_PORTS (1024-65535/tcp 1024-65535/udp)
    destination-port:: SIP 
    action:: accept
}
```
and the generated config line in the .eacl is 
```
permit udp host XYZ range 1024 65535 host ABC eq sip
```
and this is the output when you put the file on the device
```
>  permit udp host XYZ range 1024 65535 host ABC eq sip

% Invalid input at line xxx
```
However, if you replace sip with 5060, the config is accepted. 

For reference, here are the valid port aliases on my device
```
  acr-nema        ACR-NEMA Digital Imaging and Communications in Medicine (104)
  arns            A Remote Network Server System (384)
  asf-rmcp        ASF Remote Management and Control Protocol (623)
  at-rtmp         AppleTalk Routing Maintenance (201)
  aurp            Appletalk Update-Based Routing Protocol (387)
  auth            Authentication Service (113)
  bfd             Bidirectional Forwarding Detection (3784)
  bfd-echo        BFD Echo (3785)
  bftp            Background File Transfer Program (152)
  bgmp            Border Gateway Multicast Protocol (264)
  biff            Biff (mail notification, comsat) (512)
  bootpc          Bootstrap Protocol (BOOTP) client (68)
  bootps          Bootstrap Protocol (BOOTP) server (67)
  chargen         Character Generator (19)
  citadel         Citadel (504)
  clearcase       Clearcase albd (371)
  commerce        Commerce Applications (542)
  courier         Remote Procedure Call (530)
  csnet-ns        CCSO Name Server Protocol (105)
  daytime         Daytime (13)
  dhcpv6-client   DHCPv6 Client (546)
  dhcpv6-server   DHCPv6 Server (547)
  discard         Discard (9)
  dnsix           DNSIX security protocol auditing (195)
  domain          Domain Name Service (53)
[...]

PR #305, opened 2023-06-15
Fix the usage of SIP on Arista by using integer.

ankenyr:
See issue #304 
Arista does not use the word sip in place of the protocol number. This was accidentally introduced in #100 
This fix adds a simple test to check for this regression but a future test should take into account all proto to int checks.

### DIFF
--- a/tests/regression/arista/arista_test.py
+++ b/tests/regression/arista/arista_test.py
@@ -118,6 +118,13 @@ term good-term-6 {
   action:: accept
 }
 """
+GOOD_TERM_8 = """
+term good-term-2 {
+  destination-port:: SIP
+  protocol:: udp
+  action:: accept
+}
+"""
 
 SUPPORTED_TOKENS = {
     'action',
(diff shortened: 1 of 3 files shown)


tests/regression/arista/arista_test.py
```diff
@@ -313,6 +320,14 @@ def testStandardTermV4(self):
...
 
         self.naming.GetNetAddr.assert_has_calls([mock.call('SOME_HOST')])
+    
+    @capture.stdout
+    def testSIPUsesInt(self):
+        self.naming.GetServiceByProto.return_value = ['5060']
+        pol = policy.ParsePolicy(GOOD_HEADER + GOOD_TERM_8, self.naming)
+        acl = arista.Arista(pol, EXP_INFO)
+        self.assertIn('5060', str(acl))
+        self.assertNotIn('sip', str(acl))
```
jtwb, line 330:
```suggestion
        self.assertNotIn('sip', str(acl))
        print(acl)
```

Would be good to get output capture on this case.

jtwb (approved):

Issue #309, opened 2023-06-16
OpenConfig: support required sequence-id


XioNoX:
According to the spec, the sequence-id field is required at least under `acl-entry` as it's the [key](https://github.com/openconfig/public/blob/15ca44425081bbe8e32db61fc320c023a3888c42/release/models/acl/openconfig-acl.yang#L427) for this list. It's optional under "config":
```
     |     +--rw acl-entries
     |        +--rw acl-entry* [sequence-id]
     |           +--rw sequence-id        -> ../config/sequence-id
     |           +--rw config
     |           |  +--rw sequence-id?   uint32
     |           |  +--rw description?   string
```

Without it, an OpenConfig speaking device (in my case Dell Enterprise Sonic) will not accept it:
> pygnmi.client.gNMIException: GRPC ERROR Host: localhost:8080, Error: key field sequence-id (*uint32) has nil value <nil>

As a quick test I did:
```python
                        # Sequence ID
                        rule['config'] = {}
                        rule['config']['sequence-id'] = total_rule_count
                        rule['sequence-id'] = total_rule_count
```
Under https://github.com/aerleon/aerleon/blob/e44374d237d9ded3b1dda5e5a980cbb9d4161f22/aerleon/lib/openconfig.py#L263-L264 which fixed the issue. A 2nd test showed that the one under "config" is not needed.
I think it should ideally go in the Term class, but not sure yet how to do that.

As a side note, the "description" field is not far away :)

XioNoX:
Fixed in https://github.com/aerleon/aerleon/pull/312

PR #312, opened 2023-06-19
OpenConfig: add required sequence-id field

XioNoX:
Fixes issue #309

### DIFF
--- a/aerleon/lib/openconfig.py
+++ b/aerleon/lib/openconfig.py
@@ -65,7 +65,8 @@ IP = TypedDict("IP", {"config": IPConfig})
 ActionConfig = TypedDict("ActionConfig", {"forwarding-action": str})
 Action = TypedDict("Action", {"config": ActionConfig})
 ACLEntry = TypedDict(
-    "ACLEntry", {"actions": Action, "ipv4": IP, "ipv6": IP, "transport": Transport}
+    "ACLEntry",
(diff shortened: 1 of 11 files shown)


aerleon/lib/openconfig.py
```diff
@@ -262,6 +262,7 @@ def _TranslatePolicy(self, pol: Policy, exp_info: int) -> None:
                     t = Term(term, term_af)
                     for rule in t.ConvertToDict():
                         total_rule_count += 1
+                        rule['sequence-id'] = total_rule_count * 5
```
ankenyr, line 265: Why is this multiplied by 5? I am looking at the documentation at https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-acl.html and do not see anything that explains that.

XioNoX: I put the reason in the commit message.
It's sometimes (now less than before thanks to automation) useful to be able to manually add a rule in between two existing rules directly though the device's CLI. Like for testing or as incident response.

ankenyr: Ahh so for future reference, we squash commits so I don't believe those messages get preserved. You should add information like this into the initial PR.

ankenyr:
Tests are failing here. I think it is because the order of the output is wrong.

ankenyr (approved):

Issue #56, opened 2022-10-26
[LEGACY ISSUE 322] Juniper ethernet-switching for EX or QFX


ankenyr:

ProtonBruno:
This one can be closed since PR #314 and #318 addressing it were merged

PR #318, opened 2023-06-29
add ethernet-switching filter type

ProtonBruno:
Hi team,
This PR is related to issue https://github.com/aerleon/aerleon/issues/56
As discussed porting our Capirca PR here with associated unit tests.
It's basically adding `ethernet-switching` filter type for Juniper EX platform.
Thx

### DIFF
--- a/aerleon/lib/aclgenerator.py
+++ b/aerleon/lib/aclgenerator.py
@@ -103,7 +103,13 @@ class Term:
         'udplite': 136,
         'all': -1,  # Used for GCE default deny, do not use in pol file.
     }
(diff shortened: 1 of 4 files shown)


ankenyr:
This is looking good but just to ask, do we know if all features for regular filters are supported with ethernet-switching? addresses and ports look to be supported but what about advanced features such as qos, logging, etc?

ProtonBruno:
Most features are supported but indeed there are some hardware limitations. 
It's also the case for QFX filters: we had to remove logging or even filters in some scenario.
We can extend the unit tests to have more coverage and identify the gaps if you want to?

Issue #306, opened 2023-06-16
OpenConfig: add "options" support


XioNoX:
In https://github.com/aerleon/aerleon/blob/main/aerleon/lib/openconfig.py#L249 it says:
```python
                # TODO(b/196430344): Add support for options such as
                # established/rst/first-fragment
                if term.option:
                    raise OcFirewallError('OpenConfig firewall does not support term options.')
```

I don't know if OpenConfig didn't support it in the past, but seems like it does now:
https://github.com/openconfig/public/blob/15ca44425081bbe8e32db61fc320c023a3888c42/release/models/acl/openconfig-packet-match.yang#L567

It would be useful to support such usecase, especially to filter on "tcp-established" (return) traffic.

PR #320, opened 2023-07-12
OpenConfig/SONiC: add support for tcp-established option

XioNoX:
The OpenConfig implementation is experimental as only implemented based on the [OpenConfig spec](https://github.com/openconfig/public/commit/2de10a479bd0f537109108fa12bc3790aac01548) (not tested on a real device).
The SONiC implementation differs slightly from the OpenConfig one, thus the new generator inheriting from OpenConfig. This one has been tested on a real Dell SONiC device.

Fixes https://github.com/aerleon/aerleon/issues/306 and should be a decent baseline to support more options both OpenConfig and SONiC.


ankenyr:
What do you mean here by "experimental"? Is this not in sync with the current openconfig specification? I am not sure we would want to accept this into the openconfig generator if it breaks things for others.

aerleon/lib/openconfig.py
```diff
@@ -148,6 +163,17 @@ def ConvertToDict(
             protos = ['none']
 
         ace_dict = copy.deepcopy(term_dict)
+
+        # options
+        if self.term.option:
+            for opt in [str(x) for x in self.term.option]:
```
ankenyr, line 169: Are there cases when options are not a string?

ankenyr, line 169: Could this be simplified to `if self.term.protocol != ['tcp'] and 'tcp-established' in self.term.option `?
This should short circuit in the cases where the first option is false which is quick and with my other comment I don't believe options are not strings.

XioNoX: Indeed, looks like not. I copied the logic from the Juniper's generator.

XioNoX: I initially copied the code from the Juniper generator, the idea was that right now we only have 1 option, but the code needs to be ready to support more options. That said I don't think itterating over all the "options" are useful, but checking if a given option is in the array is enough (and cleaner). Let me know what you think of this new version.

XioNoX:
I updated the tests to reach 100% of code coverage.

There is something wrong going on but not sure how to fix it. It's what being caught by the tests.

If I understand correctly, the `"destination-address": "10.2.3.4/32",` is being overwritten by `"protocol": 6` instead of `"protocol": 6` to be added alongside.

Any idea what could be going on?

tests/regression/openconfig/openconfig_test.py
```diff
@@ -588,6 +604,32 @@ def testMixedDaddr(self):
         self.naming.GetNetAddr.assert_called_once_with('CORP_EXTERNAL')
         print(acl)
 
+    @capture.stdout
+    def testTcpEstablished(self):
+        self.naming.GetServiceByProto.return_value = ['80']
```
jtwb, line 609: @XioNoX I think you might have previously set a return value for GetNetAddr in this test but later removed it. I think you need to add it back. This suggested change fixes the test.

```suggestion
        self.naming.GetServiceByProto.return_value = ['80']
        self.naming.GetNetAddr.return_value = TEST_IPS
```

XioNoX: Thanks! Looks like that was it. I was looking for issues in the generator and not in the test.

ankenyr (approved):

PR #328, opened 2023-07-25
Add filter-level includes 

### LINKED ISSUE #325: Allow for include of entire policies as an include.
https://aerleon.slack.com/archives/C042MD2SLRX/p1689694063955869 for more context

The original intention was for includes to be sets of terms and used for templating. In Capirca there was no limit on the expansion and so a whole policy file could be expanded.

This unintended feature has been discovered and is in use. This however was broken in Aerleon due to not realizing it was possible.

We would like to support this as a feature.

A yaml file should allow for something like
```
filters:
  - header:
    targets:
      juniper: foo
   - include-filters: path/to/include/filter.yaml
```
where the filter.yaml file is a list of additional filters.

jtwb:
Policy files can now include other policy files at the filter level by adding an item to the filters: list like:
```yaml
filters:
-   include: "path/to/include.yaml"
```

Include items can be interspersed with filter items or other includes.

`aclgen` will try to generate ACLs for all files in the base_directory with `filters:` as the top level key, which might not be desired if a file is meant only to be included by other files. In that case, change the top level key to `filters_include_only:`, which is identical to `filters:` except that `aclgen` won't try to generate ACLs for it.

```yaml
# This file can be included by other files but will not generate ACLs directly
filters_include_only:
-   header: ...
    terms: ...
-   header: ...
    terms: ...
```

## Compatibility

This change is fully backwards compatible. Aerleon >= 1.6.0 will be required to use this feature (assuming it goes out in 1.6.0).

## Schema Update

This change extends the policy file schema. Merging this PR will automatically update JSON Schema Store and thus all IDEs that use it.

Screenshot 2023-07-26 at 9 37 31 AM
Screenshot 2023-07-26 at 9 37 22 AM

## Docs

This change updates the YAML policy reference section on includes in the docs.

Fixes #325

### DIFF
--- a/aerleon/lib/yaml.py
+++ b/aerleon/lib/yaml.py
@@ -10,15 +10,10 @@ from yaml.error import YAMLError
 
 from aerleon.lib import policy
 from aerleon.lib.policy import BadIncludePath, Policy, _SubpathOf
-from aerleon.lib.policy_builder import (
-    PolicyBuilder,
-    PolicyDict,
-    RawFilter,
-    RawFilterHeader,
-    RawPolicy,
-    RawTerm,
-)
+from aerleon.lib.policy_builder import PolicyBuilder, PolicyDict
 from aerleon.lib.yaml_loader import SpanSafeYamlLoader
+
+MAX_INCLUDE_DEPTH = 5
 
 
 class PolicyTypeError(Exception):
@@ -108,7 +103,7 @@ def ParseFile(filename, base_dir='', definitions=None, optimize=False, shade_che
             raise PolicyTypeError(
                 UserMessage("Unable to read file as YAML.", filename=filename)
             ) from yaml_error
-    policy_dict = PreprocessYAMLPolicy(filename, base_dir, policy_dict)
+    policy_dict = _preprocessYAMLPolicy(filename, base_dir, policy_dict)
     if not policy_dict:
         return
     return policy.FromBuilder(PolicyBuilder(policy_dict, definitions, optimize, shade_check))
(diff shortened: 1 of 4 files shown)


aerleon/lib/yaml.py
```diff
@@ -140,37 +137,130 @@ def ParsePolicy(
...
+                    )
+                )
+            include_path = pathlib.Path(base_dir).joinpath(filter['include'])
+            if not _SubpathOf(base_dir, include_path):
+                raise BadIncludePath(
+                    f"Include file cannot be loaded from outside the base directory. File={include_path} base_directory={base_dir}"
+                )
+            try:
+                include_file = _LoadIncludeFile(include_path)
+                include_data = yaml.load(
```
ankenyr, line 230: Why not use safe_load instead?

jtwb: The param on the next line `Loader=SpanSafeYamlLoader` causes the load to happen in safe mode.

schemas/aerleon-policies.schema.json
```diff
@@ -33,30 +46,46 @@
...
-              "type": "object"
+            "header": {
+              "type": "object",
+              "title": "Filter Header",
+              "description": "The filter header lists the generator target platforms and their options. Other filter-level settings can be defined in the header.",
+              "required": ["targets"],
+              "properties": {
+                "targets": {
+                  "title": "Filter Targets",
+                  "description": "aclgen will generate an acl for each target platforms identified here.",
```
ankenyr, line 62: s/aclgen/Aclgen?

ankenyr (approved):

schemas/aerleon-policies.schema.json
jtwb: aclgen is always lowercase because it is a command line program. We never refer to it as Aclgen even in docs or errors even at the start of sentences. But ACL should be upper case.

Issue #253, opened 2023-02-19
Juniper SRX dns-name support rendering on address-books


jbotello7381:
Juniper SRX, and maybe other platforms support to configure dns names (fqdn) in their rules.   In the case of Juniper SRX, this is supported under

```
jbotello@fwn1-lab1.jimbo.net# show | compare
[edit security zones security-zone Pub]
+     address-book {
+         address DNS {
+             dns-name myserver.jimbo.net;
+         }
+     }

[edit]
```

Currently, aerleon definitions->networks doesn't support address in form of fqdn.  

It seems the use of verbatim is not possible either, since is only supported on terms.  

In order to improve the range of aerleon Juniper SRX support, it would be important for aerleon support dns-names in the networks definitions.

I'm not sure what option would be more aligned to the aerleon design.

```
networks:
   MYNET:
        values:
             - address: 10.10.10.10
             - address: myserver.jimbo.net
```

or

```
networks:
   MYNET:
        values:
             - address: 10.10.10.10
             - dns-name: myserver.jimbo.net
```
or any other key name that could be more aligned to other platforms in general? for example, Palo Alto i believe it define it as `fqdn`

```
networks:
   MYNET:
        values:
             - address: 10.10.10.10
             - fqdn: myserver.jimbo.net
```

jtwb:
Question: does Juniper SRX accept any host name or is a fully qualified name required? Also, does PA Networks actually require an a fully qualified name (per the field "fqdn") or will it accept any host name? If the firewall will resolve these host names against internal DNS, they might not be fully qualified, which is why I suspect any host name might be accepted. My preference would be for the name for this new field to accurately match the type of data it accepts, so if any hostname is accepted, "hostname" or "host" might be the most accurate term.

What should happen with generators that do not support dns-name, either because the target platform does not support it or because the generator was written before dns-name was added to Aerleon? Generators usually are written to bail out if they encounter a policy they cannot safely process, and that would be a reasonable behavior here, except we are adding this feature to a relatively mature ecosystem where we know companies have their own private generators which we can neither see nor control. If those private generators are written in such a way that they crash when encountering networks with dns-name data, that would actually be a pretty OK outcome. The worst case would be if a private generator silently ignores the dns-name data and happily generates an ACL with that critical data missing.

[...]

PR #330, opened 2023-08-02
Adding FQDN filtering into Juniper SRX generator.

ankenyr:
This completes the work to resolve https://github.com/aerleon/aerleon/issues/253
SRX will now be able to filter FQDN's in their terms.

FQDN's can be used in a term by using either `destination-fqdn` or `source-fqdn` the below is an example

```
  - name:  allow_dns
    destination-fqdn: CLOUDFLARE_DNS
    action: accept
```

See https://github.com/aerleon/aerleon/pull/291 for how to add FQDN's to the addressbook.

### DIFF
--- a/aerleon/lib/addressbook.py
+++ b/aerleon/lib/addressbook.py
@@ -12,19 +12,29 @@ class Addressbook:
     def __init__(self) -> None:
         self._fqdn_addressbook = collections.OrderedDict()
         self._ip_addressbook = collections.OrderedDict()
+        
+    def GetZoneNames(self) -> List:
+        return list(self._ip_addressbook.keys())
+    
+    def WalkAddressBook(self, zone_filter=None):
+        
+        return self._walk(zone_filter)
 
-    def WalkAddressBook(self):
-        return self._walk()
-    def _walk(self):
+    def _walk(self, zone_filter=None):
         for zone in self._ip_addressbook:
-            groups = self._ip_addressbook[zone].keys()
-            groups.extend(self._fqdn_addressbook[zone].keys())
-            for group in groups:
+            if zone_filter and zone != zone_filter:
+                continue
+            groups = set(self._ip_addressbook[zone].keys())
+            groups.update(list(self._fqdn_addressbook[zone].keys()))
+            for group in sorted(groups):
                 ips = []
+                fqdns = []
                 if group in self._ip_addressbook[zone]:
                     ips = self._ip_addressbook[zone][group]
-                yield zone, group
-        
+                if group in self._fqdn_addressbook[zone]:
+                    fqdns = self._fqdn_addressbook[zone][group]
+                yield zone, group, ips, fqdns
+
     def GetFQDN(self, zone: str, name: str):
         return self._fqdn_addressbook[zone][name]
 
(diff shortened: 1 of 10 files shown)


aerleon/lib/addressbook.py
```diff
@@ -4,12 +4,55 @@
...
+                yield zone, group, ips, fqdns
+
+    def GetFQDN(self, zone: str, name: str):
+        return self._fqdn_addressbook[zone][name]
+
+    def GetAddressTokensInZone(self, zone: str):
+        return [i for i in self._ip_addressbook['zone']]
+
+    def GetFQDNTokensInZone(self, zone: str):
+        return [i for i in self._fqdn_addressbook['zone']]
```
jtwb, line 45: Looks like a typo. There should probably be a quick test that covers these functions.

```suggestion
    def GetAddressTokensInZone(self, zone: str):
        return [i for i in self._ip_addressbook[zone]]

    def GetFQDNTokensInZone(self, zone: str):
        return [i for i in self._fqdn_addressbook[zone]]
```

jtwb: Also, could this just be `return copy(self._fqdn_addressbook[zone])` ?

aerleon/lib/junipersrx.py
```diff
@@ -529,8 +537,8 @@ def _TranslatePolicy(self, pol: policy.Policy, exp_info: int):
                 # SRX policies are controlled by addresses that are used within, so
                 # policy can be at the same time inet and inet6.
                 if self._GLOBAL_ADDR_BOOK in self.addr_book_type:
-                    for zone in self.addressbook.addressbook:
-                        for _, ips in sorted(self.addressbook.addressbook[zone].items()):
+                    for zone in self.addressbook._ip_addressbook:
+                        for _, ips in sorted(self.addressbook._ip_addressbook[zone].items()):
```
jtwb, line 541: Should this code use one of the new addressbook methods?

aerleon/lib/cisco.py
```diff
@@ -208,9 +208,10 @@ def __str__(self) -> str:
         # for using cisco, which has decided to implement its own meta language.
 
         # Create network object-groups
-        for name, ips in self.addressbook.addressbook[''].items():
+        for zone, name, ips, _ in self.addressbook.WalkAddressBook():
             for version in (4, 6):
                 vips = [i for i in ips if i.version == version]
+                vips = nacaddr.CollapseAddrList(vips)
```
jtwb, line 214: Is the addition of CollapseAddrList here related to the FQDN / AddressBook change?

aerleon/lib/addressbook.py
```diff
@@ -4,12 +4,55 @@
...
 class Addressbook:
     def __init__(self) -> None:
-        self.addressbook = collections.OrderedDict()
+        self._fqdn_addressbook = collections.OrderedDict()
+        self._ip_addressbook = collections.OrderedDict()
+
+    def GetZoneNames(self) -> List:
+        return list(self._ip_addressbook.keys())
+
+    def WalkAddressBook(self, zone_filter=None):
```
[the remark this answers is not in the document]
ankenyr: Done

aerleon/lib/addressbook.py
ankenyr: Done

aerleon/lib/junipersrx.py
ankenyr: naaaa

jtwb (approved):

PR #333, opened 2023-08-04
Add support for Nokia SR Linux

jbemmel:
Basic ACL support for the following features:
* IPv4/v6 source/dest IP and ports
* IP&TCP flags

Includes regression test set for most of the above ( IP&TCP flags TBD )


aerleon/lib/nokiasrl.py
```diff
@@ -0,0 +1,303 @@
+# Copyright 2021 Google Inc. All Rights Reserved.
```
ankenyr, line 1: You can remove the copyright notices if this was created by yourself.
You can see in any code that was not forked from Capirca we have left the header off.

ankenyr:
Looks like the formatter failed. You will want to run black and isort. You can do this by running `nox -s format` while inside a poetry shell.

aerleon/lib/nokiasrl.py
jbemmel: I mostly copied from the openconfig file, so I figured I'd leave it in place

ankenyr: I actually just noticed this. There is a lot of overlap here. What happens if we find a bug in the openconfig code? We will have to fix it in two places. That isn't a great position to be in. Would you be able to inherit from Openconfig or create a common class for them to inherit from?

ankenyr:
Does Nokia have any documentation we can reference for now and in the future as we maintain the generator?

aerleon/lib/nokiasrl.py
jbemmel: Sure, I suppose we could create a common base class and then have generator methods for source-address, dest-address, etc. ( i.e. refactor Openconfig and then derive SRLinux from that, overriding those methods )

jbemmel:
> Does Nokia have any documentation we can reference for now and in the future as we maintain the generator?

See the comment on line 19: More information about the SR Linux ACL model schema: https://yang.srlinux.dev/

jbemmel:
Refactoring done, please review

aerleon/lib/nokiasrl.py
```diff
@@ -0,0 +1,162 @@
+# Copyright 2023 Nokia All Rights Reserved.
```
ankenyr, line 1: Just want to check on this since you are contributing with your personal gmail account. You are wishing to provide copyright to Nokia and you can attest that the code made was created by you on behalf of Nokia?

You can also add the company to the `AUTHORS` file in the root directory if you desire.

tests/regression/nokiasrl/nokiasrl_test.py
```diff
@@ -0,0 +1,434 @@
+# Copyright 2021 Google Inc. All Rights Reserved.
```
ankenyr, line 1: Might want to fix this copyright also

aerleon/lib/openconfig.py
```diff
@@ -70,12 +70,14 @@ def RecursiveDict() -> DefaultDict[Any, Any]:
 )
 aclEntries = TypedDict("aclEntries", {"acl-entry": List[ACLEntry]})
 ACLSetConfig = TypedDict("ACLSetConfig", {"name": str, "type": str})
-ACLSet = TypedDict(
-    "ACLSet", {"acl-entries": aclEntries, "config": ACLSetConfig, "name": str, "type": str}
-)
 
 
-class Term(aclgenerator.Term):
+# ACLSet = TypedDict(
```
ankenyr, line 75: Why the comment out of ACLSet here?

ankenyr:
Made a few comments while I had a few moments.

aerleon/lib/nokiasrl.py
jbemmel: Yes, copyright belongs to Nokia and I created the code on its behalf

Added the company name to AUTHORS

tests/regression/nokiasrl/nokiasrl_test.py
jbemmel: done

aerleon/lib/openconfig.py
jbemmel: Because it's no longer being used - with the rewrite, the dict can contain either openconfig or srlinux ACL entries (different structure)

I'm not sure if/how validation works, I didn't find a difference in having these specs be correct or not - may need further review

tests/regression/nokiasrl/nokiasrl_test.py
```diff
@@ -0,0 +1,434 @@
...
+
+    @capture.stdout
+    def testSaddr(self):
+        self.naming.GetNetAddr.return_value = TEST_IPS
+
+        acl = nokiasrl.NokiaSRLinux(
+            policy.ParsePolicy(GOOD_HEADER + GOOD_SADDR, self.naming), EXP_INFO
+        )
+        expected = json.loads(GOOD_JSON_SADDR)
+        self.assertEqual(expected, json.loads(str(acl)))
```
ankenyr, line 322: The output is already checked with @capture.stdout so this assertion isn't strictly necessary imo.

tests/regression/nokiasrl/nokiasrl_test.py
```diff
@@ -0,0 +1,434 @@
...
+_TERM_TARGET_TAGS_LIMIT = 70
+_TERM_PORTS_LIMIT = 256
+
+
+class NokiaSRLTest(absltest.TestCase):
+    def setUp(self):
+        super().setUp()
+        self.naming = mock.create_autospec(naming.Naming)
+
+    def _StripAclHeaders(self, acl):
```
ankenyr, line 309: Is this used? I am missing seeing where it is.

aerleon/lib/nokiasrl.py
```diff
@@ -0,0 +1,162 @@
...
+    ACTION_MAP = {'accept': 'accept', 'deny': 'drop', 'reject': 'drop'}
+
+    def SetAction(self, dict: dict) -> None:
+        action = self.ACTION_MAP[self.term.action[0]]
+        dict['action'] = {action: {}}
+
+    def SetOptions(self, dict: dict) -> None:
+        # Handle various options
+        opts = [str(x) for x in self.term.option]
+        _match = dict['match'] = {}
```
ankenyr, line 72: Why the double assignment here? You use `dict['match']` in other sections.

tests/regression/nokiasrl/nokiasrl_test.py
jbemmel: It's not, I copied it from the OpenConfig test code. Removed in both

aerleon/lib/nokiasrl.py
jbemmel: ok, made consistent

tests/regression/nokiasrl/nokiasrl_test.py
jbemmel: It's superfluous, I copied it from the OpenConfig tests. It's in all the methods - also in other parts.

e.g. https://github.com/aerleon/aerleon/blob/main/tests/regression/cisconx/cisconx_test.py#L239
is also checked by stdout capture.

Happy to remove them, but it may be better to do that for all tests in a separate commit/branch

aerleon/lib/openconfig.py
```diff
@@ -196,17 +187,52 @@ def ConvertToDict(
...
         return rules
 
+    def SetAction(self, dict: dict) -> None:
+        action = self.ACTION_MAP[self.term.action[0]]
+        dict['actions'] = {}
+        dict['actions']['config'] = {}
+        dict['actions']['config']['forwarding-action'] = action
+
+    def SetOptions(self, dict: dict) -> None:
+        pass  # no-op, override in subclasses
```
ankenyr, line 207: This comment is not needed.

aerleon/lib/openconfig.py
ankenyr: The specs are used for type hinting. https://peps.python.org/pep-0484/
Eventually once we have things fully type hinted we can use mypy or some other tool for static type analysis which will give us more confidence in code correctness.

ankenyr: You will want to try and get them things right for the different modules. You should be able to overwrite the ACLSet used in nokiasrl with a Type of the correct structure.

aerleon/lib/openconfig.py
```diff
@@ -196,17 +187,52 @@ def ConvertToDict(
...
-                                ace_dict[family]['config']['protocol'] = proto_num
+                                self.SetProtocol(ace_dict, family, proto)
                                 # This is the business end of ace explosion.
-                                # A dict is a reference type, so deepcopy is atually required.
+                                # A dict is a reference type, so deepcopy is actually required.
                                 rules.append(copy.deepcopy(ace_dict))
 
         return rules
 
+    def SetAction(self, dict: dict) -> None:
```
ankenyr, line 200: A lot of these functions are being called `SetFoo` but are not setting something on the class but on some passed in object. SetAction is likely not even a public function. If someone makes a Openconfig object are they expected to ever call SetAction? You may wish to consider assigning the dict to the object and operating it itself.

aerleon/lib/nokiasrl.py
```diff
@@ -0,0 +1,162 @@
...
+        Returns:
+          tuple containing both supported tokens and sub tokens
+        """
+        supported_tokens, supported_sub_tokens = super()._BuildTokens()
+
+        # Remove unsupported things, note icmp-type could be supported
+        supported_tokens -= {'platform', 'platform_exclude', 'verbatim', 'icmp-type'}
+
+        # SR Linux ACL model only supports these 2 forwarding actions.
+        supported_sub_tokens['action'] = {'accept', 'deny'}  # removed 'reject'
```
ankenyr, line 127: You can remove the comments here and in the supported_sub_tokens section.

ankenyr:
You will also want to add documentation to the generators page.
https://github.com/aerleon/aerleon/blob/main/docs/reference/generators.md
https://aerleon.readthedocs.io/en/latest/reference/generators/

aerleon/lib/openconfig.py
jbemmel: added ```term_dict``` as a member

ankenyr (approved):

PR #334, opened 2023-08-05
Update usage of icmpv6 or icmp6 in Juniper.

### LINKED ISSUE #323: Minor regression on juniper generator
After upgrading from Capirca to Aerleon, the generated `juniper` policies use `icmpv6` instead of `icmp6`
```diff
[edit firewall family inet6 filter loopback6 term allow_ok_icmp6 from]
-       next-header icmp6;
+       next-header icmpv6;
```

Both work but according to Juniper's [doc](https://www.juniper.net/documentation/us/en/software/junos/routing-policy/topics/concept/firewall-filter-match-conditions-for-ipv6-traffic.html):
> Note: next-header icmp6 and next-header icmpv6 match conditions perform the same function. next-header icmp6 is the preferred option. next-header icmpv6 is hidden in the Junos OS CLI.

So there is a risk that `icmpv6` becomes decommissioned at some point and breaks.

ankenyr:
This fixes #323

Juniper has changed their syntax to prefer icmp6 over icmpv6. Our YAML language is agnostic of these changes but our generators should be updated to use what ever the platform prefers. This change adds a method that looks for the usage of icmpv6 in the protocol and replaces it with icmp6 for output.

@XioNoX for visibility

### DIFF
--- a/aerleon/lib/juniper.py
+++ b/aerleon/lib/juniper.py
@@ -231,7 +231,7 @@ class Term(aclgenerator.Term):
         # Don't render icmpv6 protocol terms under inet, or icmp under inet6
         if (self.term_type == 'inet6' and 'icmp' in self.term.protocol) or (
             self.term_type == 'inet'
-            and ('icmpv6' in self.term.protocol or 'icmpv6' in self.term.protocol)
+            and ('icmpv6' in self.term.protocol or 'icmp6' in self.term.protocol)
         ):
             logging.warning(
                 self.NO_AF_LOG_PROTO.substitute(
--- a/aerleon/lib/juniper.py
+++ b/aerleon/lib/juniper.py
@@ -16,7 +16,7 @@
 
 """Juniper JCL generator."""
 
-
+import copy
 from typing import Dict, List, Set, Tuple, Union
 
 from absl import logging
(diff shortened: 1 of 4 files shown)


tests/regression/juniper/juniper_test.py
```diff
@@ -1009,13 +1009,13 @@ def testProtocolExcept(self):
         self.assertIn('next-header-except tcp;', output, output)
         print(output)
 
-    @capture.stdout
+    # @capture.stdout
```
jtwb, line 1012: Was this test accidentally disabled?

jtwb: Or rather the capture assertion?

aerleon/lib/juniper.py
```diff
@@ -781,6 +782,19 @@ def _MinimizePrefixes(
 
         return include_result, exclude_result
 
+    def _ReplaceICMPv6(self):
+        """Juniper has deprecated the use ic icmpv6 in favor of icmp6.
```
jtwb, line 786:
```suggestion
        """Juniper has deprecated the use of icmpv6 in favor of icmp6.
```

aerleon/lib/juniper.py
```diff
@@ -781,6 +782,19 @@ def _MinimizePrefixes(
...
+        https://github.com/aerleon/aerleon/issues/323
+        """
+        if 'icmpv6' in self.term.protocol:
+            i = self.term.protocol.index('icmpv6')
+            del self.term.protocol[i]
+            self.term.protocol.insert(i, 'icmp6')
```
jtwb, line 792:
```suggestion
            self.term.protocol[i] = 'icmp6'
```

aerleon/lib/juniper.py
```diff
@@ -781,6 +782,19 @@ def _MinimizePrefixes(
...
+        if 'icmpv6' in self.term.protocol_except:
+            i = self.term.protocol_except.index('icmpv6')
+            del self.term.protocol_except[i]
+            self.term.protocol_except.insert(i, 'icmp6')
```
jtwb, line 796:
```suggestion
            self.term.protocol_except[i] = 'icmp6'
```

aerleon/lib/juniper.py
jtwb: Also the index variable is named `loc` elsewhere in `Term.__init__`, may as well name it that, as `i` is conventionally used for numeric loop counters.

aerleon/lib/juniper.py
```diff
@@ -224,6 +224,7 @@ def __init__(
 
             # some options need to modify the actions
             self.extra_actions = []
+        self._ReplaceICMPv6()
```
jtwb, line 227: Crazy suggestion: does this need to be a separate method? It looks like lines 218-223 do something very similar here in `Term.__init__`, so it's probably fine to go with the flow and have the code live right here.

tests/regression/juniper/juniper_test.py
ankenyr: It was, those need to be disabled if you want to drop into a debugger.

aerleon/lib/juniper.py
ankenyr: Done

ankenyr: the change from i to loc is sane but why the suggestion of `self.term.protocol[i] = 'icmp6'`?
If the list is empty it will return an `IndexError` and if it is not we are overwriting a valid entry

```
>>> foo = ['a','b']
>>> loc = foo.index('a')
>>> del foo[loc]
>>> foo
['b']
>>> foo[loc] = 'a'
>>> foo
['a']
```

jtwb: The list cannot be empty here because you are guarding with `if 'icmpv6' in self.term.protocol:`.

ankenyr: d'oh

jtwb (approved):

PR #335, opened 2023-08-07
Adding packet/byte count to IPTables restore format (speedway)

### LINKED ISSUE #321: Speedway does not include packet/byte count in its output
Normally the iptables-save command will provide details on how many packets and bytes have gone through the specific filter. The syntax is like 

```
:OUTPUT DROP [0:0]
-N O_deny-all
-A O_deny-all -p tcp -j DROP
-A O_deny-all -p udp -j DROP
COMMIT
```
where the counters are specifically `[0:0]`

This is not included in our output. This may cause some unknown incompatibility we have not heard of or it might be fine. In either case any parser that is working off outputs from `iptables-save` will expect to see this in our output.

ankenyr:
This fixes #321 
The packet/byte count is required for IPTables restore.

### DIFF
--- a/tests/regression/speedway/speedway_test.py
+++ b/tests/regression/speedway/speedway_test.py
@@ -163,16 +163,16 @@ class SpeedwayTest(absltest.TestCase):
             result[0],
             '*filter designation does not appear at top of generated ' 'policy.',
         )
-        # self.assertIn(':INPUT ACCEPT [0:0]', result, 'input default policy of accept not set.')
-        # self.assertIn('-N I_good-term-1', result, 'did not find new chain for good-term-1.')
-        # self.assertIn(
-        #     '-A I_good-term-1 -p icmp -m state --state NEW,ESTABLISHED,RELATED' ' -j ACCEPT',
-        #     result,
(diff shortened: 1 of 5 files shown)


tests/regression/speedway/speedway_test.py
```diff
@@ -163,16 +163,16 @@ def testSpeedwayOutputFormat(self):
...
-        self.assertIn('-N I_good-term-1', result, 'did not find new chain for good-term-1.')
-        self.assertIn(
-            '-A I_good-term-1 -p icmp -m state --state NEW,ESTABLISHED,RELATED' ' -j ACCEPT',
-            result,
-            'did not find append for good-term-1.',
-        )
-        self.assertEqual(
-            'COMMIT', result[len(result) - 2], 'COMMIT does not appear at end of output policy.'
-        )
+        # self.assertIn(':INPUT ACCEPT [0:0]', result, 'input default policy of accept not set.')
```
jtwb, line 166: Intentionally commented out?

ankenyr: That was unintended. fixed.

jtwb (approved):

PR #339, opened 2023-08-11
Add Nokia SR Linux support for established and tcp-established options, and comments

jbemmel:
* Also fixes bug to set 'tcp' protocol when tcp-flags are used
* Also sets 'name' attribute for policies

### DIFF
--- a/aerleon/lib/nokiasrl.py
+++ b/aerleon/lib/nokiasrl.py
@@ -89,7 +89,7 @@ class SRLTerm(openconfig.Term):
                 log = {"log": True}
             else:
                 raise UnsupportedLogging(
-                    'logging can only be used with deny in term %s' % self.term.name
+                    f'logging can only be used with deny in term {self.term.name}'
                 )
         self.term_dict['action'] = {action: log}
 
@@ -122,8 +122,7 @@ class SRLTerm(openconfig.Term):
                 _tcp_established()
             else:
                 raise TcpEstablishedWithNonTcpError(
-                    'tcp-established can only be used with tcp protocol in term %s'
-                    % self.term.name
+                    f'tcp-established can only be used with tcp protocol in term {self.term.name}'
                 )
         elif 'established' in opts:
             if self.term.protocol:
@@ -135,12 +134,11 @@ class SRLTerm(openconfig.Term):
                         self.SetDestPorts(1024, 65535)
                 else:  # Could produce 2 rules if [tcp,udp]
                     raise EstablishedWithNonTcpUdpError(
-                        'established can only be used with tcp or udp protocol in term %s'
-                        % self.term.name
+                        f'established can only be used with tcp or udp protocol in term {self.term.name}'
                     )
             else:
                 raise EstablishedWithNoProtocolError(
-                    'must specify a protocol for "established" in term %s' % self.term.name
+                    f'must specify a protocol for "established" in term {self.term.name}'
                 )
 
         if 'tcp-flags' in self.term_dict['match']:
--- a/aerleon/lib/nokiasrl.py
+++ b/aerleon/lib/nokiasrl.py
@@ -52,7 +52,8 @@ ACLEntry = TypedDict(
     "ACLEntry",
     {"sequence-id": int, "action": Action, "match": Match},
 )
-IPFilters = TypedDict("IPFilters", {"ipv4-filter": List[ACLEntry], "ipv6-filter": List[ACLEntry]})
+Entries = TypedDict("Entries", {"entry": List[ACLEntry], "description": str})
+IPFilters = TypedDict("IPFilters", {"ipv4-filter": Entries, "ipv6-filter": Entries})
 
 # generic error class
 class Error(aclgenerator.Error):
@@ -79,6 +80,9 @@ class SRLTerm(openconfig.OCTerm):
     def SetAction(self) -> None:
         action = self.ACTION_MAP[self.term.action[0]]
         self.term_dict['action'] = {action: {}}
+
+    def SetComments(self, comments: List[str]) -> None:
+        self.term_dict['description'] = "\n".join(comments)
 
     def SetOptions(self, family: str) -> None:
         # Handle various options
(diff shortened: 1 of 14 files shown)


aerleon/lib/nokiasrl.py
```diff
@@ -62,9 +83,20 @@ class SRLTerm(openconfig.OCTerm):
...
+                raise UnsupportedLogging(
+                    'logging can only be used with deny in term %s' % self.term.name
+                )
+        self.term_dict['action'] = {action: log}
 
-    def SetOptions(self) -> None:
+    def SetComments(self, comments: List[str]) -> None:
+        self.term_dict['description'] = "\n".join(comments)
+
+    def SetOptions(self, family: str) -> None:
```
ankenyr, line 99: Family looks to be unused.

aerleon/lib/nokiasrl.py
```diff
@@ -62,9 +83,20 @@ class SRLTerm(openconfig.OCTerm):
 
     def SetAction(self) -> None:
         action = self.ACTION_MAP[self.term.action[0]]
-        self.term_dict['action'] = {action: {}}
+        log = {}
+        if self.term.logging:
+            if action == 'drop':
+                log = {"log": True}
+            else:
+                raise UnsupportedLogging(
```
ankenyr, line 91: Can we make a test for this case?

aerleon/lib/nokiasrl.py
```diff
@@ -62,9 +83,20 @@ class SRLTerm(openconfig.OCTerm):
...
+                    'logging can only be used with deny in term %s' % self.term.name
```
ankenyr, line 92: You should try and use f-strings, they are much more readable. This would be f'logging can only be used with deny in term {self.term.name}'

aerleon/lib/nokiasrl.py
```diff
@@ -81,7 +113,38 @@ def SetOptions(self) -> None:
             )
         if 'not-syn-ack' in opts:
             self.term_dict['match']['tcp-flags'] = "!(syn&ack)"
-        # Note: not handling established | tcp-established, could throw error
+
+        def _tcp_established():
+            self.term_dict['match']['tcp-flags'] = "ack|rst"
```
ankenyr, line 118: This seems wrong to me, wouldn't tcp established be syn and ack?

aerleon/lib/nokiasrl.py
```diff
@@ -81,7 +113,38 @@ def SetOptions(self) -> None:
...
+                    % self.term.name
+                )
+        elif 'established' in opts:
+            if self.term.protocol:
+                if self.term.protocol == ['tcp']:
+                    _tcp_established()
+                elif self.term.protocol == ['udp']:
+                    self.SetProtocol(family=family, protocol="udp")
+                    if not self.term.destination_port:
+                        self.SetDestPorts(1024, 65535)
```
ankenyr, line 135: Why add specific ports when none were specified?

aerleon/lib/nokiasrl.py
```diff
@@ -81,7 +113,38 @@ def SetOptions(self) -> None:
...
+                else:  # Could produce 2 rules if [tcp,udp]
+                    raise EstablishedWithNonTcpUdpError(
```
ankenyr, line 137: Add a test case for this error please.

aerleon/lib/nokiasrl.py
```diff
@@ -81,7 +113,38 @@ def SetOptions(self) -> None:
...
+                        'established can only be used with tcp or udp protocol in term %s'
+                        % self.term.name
+                    )
+            else:
+                raise EstablishedWithNoProtocolError(
```
ankenyr, line 142: This one too please

aerleon/lib/nokiasrl.py
```diff
@@ -124,12 +187,12 @@ def _BuildTokens(self) -> Tuple[Set[str], Dict[str, Set[str]]]:
 
         supported_sub_tokens['action'] = {'accept', 'deny'}  # excludes 'reject'
         supported_sub_tokens['option'] = {
-            #  'established',
+            'established',
             'first-fragment',
             'is-fragment',
             'fragments',
             #  'sample',
```
ankenyr, line 194: Are the commented out sub tokens to be added later? If not can we remove those?

tests/regression/nokiasrl/nokiasrl_test.py
```diff
@@ -417,6 +481,62 @@ def testV6Daddr(self):
...
+    def testEstablishedNonTcpUdp(self):
+        self.naming.GetServiceByProto.return_value = ['53']
+
+        acl = policy.ParsePolicy(GOOD_HEADER + BAD_TERM_2, self.naming)
+        with self.assertRaises(nokiasrl.TcpEstablishedWithNonTcpError):
+            _ = nokiasrl.NokiaSRLinux(acl, EXP_INFO)
+
 
 #
 # TODO:
```
ankenyr, line 542: Think you need to remove from the TODO's down here?

ankenyr:
Made a first pass, this is looking good though. You will need to resolve the conflicts, let me know if you need help with that.

aerleon/lib/nokiasrl.py
jbemmel: See https://github.com/aerleon/aerleon/blob/main/docs/reference/generators.md (many vendors)

e.g.
```
Option
option: {established|tcp-established|initial|rst|first-fragment}
established: Only match established connections, implements tcp-established for tcp and sets destination port to 1024-65535 for udp if destination port is not defined.
```

aerleon/lib/nokiasrl.py
jbemmel: See https://github.com/aerleon/aerleon/blob/main/aerleon/lib/juniper.py#L182

Based on customer feedback

aerleon/lib/nokiasrl.py
jbemmel: It is used in 
```self.SetProtocol(family=family, protocol="udp")```

Needed because the attribute name is different for ipv4 versus ipv6

aerleon/lib/nokiasrl.py
jbemmel: done

aerleon/lib/nokiasrl.py
jbemmel: done

aerleon/lib/nokiasrl.py
jbemmel: leaving as-is for now

ankenyr (approved):

PR #344, opened 2023-08-18
Add revision and other tags as annotation

jbemmel:


aerleon/lib/nokiasrl.py
```diff
@@ -52,9 +52,12 @@
     "ACLEntry",
     {"sequence-id": int, "action": Action, "match": Match},
 )
-Entries = TypedDict("Entries", {"entry": List[ACLEntry], "description": str})
+Entries = TypedDict(
+    "Entries", {"entry": List[ACLEntry], "description": str, "name": str, "_annotate": str}
```
ankenyr, line 56: You added name but did you want to add the tests for them?

jbemmel: It's covered by the updated .ref files, still not sure how the TypedDict would be tested/enforced

ankenyr (approved):

PR #355, opened 2023-10-31
Fix typos, style and toc

btriller:
* Fix some typos, style and NSXt/v docs.
* Drop some generated sample policies.
* This also fixes missing generator links in TOC


docs/reference/generators.md
```diff
@@ -1148,62 +1145,42 @@ targets:
 
 ***
 
-# NSXT
+## NSXt
```
ankenyr, line 1148: Why this change?

ankenyr: Most cases I see have all letters capitalized. https://www.vmware.com/products/nsx/migrate-to-nsx-t.html

btriller: You're right. Do you mind a dash between NSX and T/v?

ankenyr: I don't mind a dash in documentation and comments. Don't change any modules, classes, or method names.

ankenyr (approved):

PR #366, opened 2024-04-09
Updated for new R24.3 ACL syntax, add per-ACL stats

jbemmel:
* Generate per-ACL stats by default, add option 'nostats' to disable
* add test and update documentation


ankenyr:
Thanks for the PR. Can you add tests to ensure that the feature `pre2024` does what it is supposed to?

PR #367, opened 2024-04-18
Changing Nokia options from pre2024 to post2024 and nostats to filterstats inverting the defaults.

ankenyr:
Upon further review we decided to invert the options created in #366. This will preserve the current functionality for current users. As people upgrade their Nokia platforms they can migrate to using `filterstats` and `post2024` options.
@jbemmel and @ProtonBruno for visibility


jtwb (approved):

ProtonBruno:
Hi,
It sounds convenient for now but may become less in the future and may require another invert imho.
Anyway thank you for the follow up.

jbemmel:
I am not aware of many customers using this at the moment, I think most customers adopting this will likely be using SR Linux R24.3 or higher. Making them put a magic flag to obtain the correct format is unnecessarily complicating, imho

Both options were added in consultation with the only customer that I am aware of using this right now - Proton. They indicated their preference is to have stats enabled by default.

It is your project and I can accept whatever the community decides is best - but my preference would be to revert this change

Issue #383, opened 2025-01-26
Notice: Upcoming Syntax Change in Aerleon for SR Linux Compatibility


ankenyr:
Nokia's SR Linux version r24.3.2, released in March 2024, switched to a new ACL syntax which is not compatible with earlier versions. To ease the migration process, these changes were initially placed behind an optional setting, allowing users to upgrade at their own pace.

Aerleon is now planning to invert these options. Starting with the first release in March, the default behavior will align with the updated syntax from SR Linux r24.3.2.

Please see [this comparison of the old and new syntax for an example ACL](https://github.com/aerleon/aerleon/pull/366/files#diff-f84f6f691314f02a5b006a5ecb233d73cd0ee1ff184a27e2df6a11b6ce907901) for more detail.

Key Changes:

* Users who have not yet upgraded to the new syntax will need to explicitly enable an option to retain compatibility with the pre-r24.3.2 syntax starting in March.
* A warning will be added in the code to help users identify and prepare for this transition.

We encourage all users to review their configurations and ensure they are ready for this change. This update is part of our commitment to maintaining compatibility and supporting the latest industry standards.

PR #384, opened 2025-01-26
Adding warning for nokia defaults changing.

ankenyr:
Nokia has changed their syntax on their SRLinux platform starting with r24.3.2. Currently our code outputs by default the old syntax and the new syntax is locked behind a filter option. This new behavior has been out for some time and the default behavior should upgrade with our user base. In March we will invert the options and the default behavior will be the new syntax.

Until then we want to provide notice to anyone using Nokia that these changes are coming so they can be prepared. This change introduces a warning that prints out anytime Nokia is used regardless so all users are notified that this change is incoming.

See https://github.com/aerleon/aerleon/issues/383 for announcement.

### DIFF
--- a/aerleon/lib/nokiasrl.py
+++ b/aerleon/lib/nokiasrl.py
@@ -22,6 +22,8 @@ More information about the SR Linux ACL model schema: https://yang.srlinux.dev/
 import sys
 from typing import Dict, List, Set, Tuple
 
+from absl import logging
+
 from aerleon.lib import aclgenerator, openconfig
 from aerleon.lib.policy import Term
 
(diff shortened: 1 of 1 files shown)


aerleon/lib/nokiasrl.py
```diff
@@ -247,6 +249,9 @@ def _TranslateTerms(
     ) -> None:
         srl_acl_entries: Dict[str, List[ACLEntry]] = {'inet': [], 'inet6': []}
         afs = ['inet', 'inet6'] if address_family == 'mixed' else [address_family]
+        logging.warning(
```
jtwb, line 252: Is this message going to appear once per input file? Could the Python 'warning' library be used to suppress repetition of the warning?

https://docs.python.org/3/library/warnings.html#:~:text=Repetitions%20of%20a%20particular%20warning%20for%20the%20same%20source%20location%20are%20typically%20suppressed.

aerleon/lib/nokiasrl.py
```diff
@@ -247,6 +249,9 @@ def _TranslateTerms(
...
+            "NOTICE: Default behavior for Nokia SRLinux will change in March 2025. Please see https://github.com/aerleon/aerleon/issues/383"
```
jtwb, line 253:
```suggestion
            "NOTICE: The Nokia SRLinux default ACL output will switch to the r24.3 syntax in Aerleon 1.11 . Please see https://github.com/aerleon/aerleon/issues/383"
```

aerleon/lib/nokiasrl.py
jtwb: To put the repetition in perspective, I think @ProtonBruno has 200+ input files. Can we also skip this message if they are already using the r24.3 option?

PR #386, opened 2025-01-28
Fix wrong Log statement on SRLinux release > 24.3.2

ProtonBruno:
Hi,
We found out that the new SRLinux r24.3.2 syntax rendered by Aerleon was wrong.
This PR fixes that.

The current wrong syntax:
`          "action": {
            "drop": {
              "log": true
            }
          },`

The correct syntax is:
`         "action": {
            "drop": {},
            "log": true
          },`

Thx.


ankenyr:
@all-contributors add @ProtonBruno for code

PR #396, opened 2025-05-22
Proxmox Firewall generator

loulecrivain:
This PR adds support for generating Proxmox Firewall configuration files.

Proxmox Virtual Environment (Proxmox VE Hypervisor) includes firewall capabilities at the cluster, host, "VNet" (aka virtual network) and virtual machine levels. See [Proxmox Firewall documentation](https://pve.proxmox.com/pve-docs/chapter-pve-firewall.html) for more information. Please note that even while it uses iptables or nftables underneath, I have chosen to not derive from aerleon existing code since the Proxmox Firewall syntax is far more simplified and generic.

The output has been tested against Proxmox VE Hypervisor version `pve-manager/8.4.1/2a5fa54a8503f96d (running kernel: 6.8.12-10-pve)`. 

There is some stuff that I'm not 100% confident about "if we should do it like this" (more specifically the tens of options for the filter, but I wanted something that could directly generate the files). So feel free to suggest changes.


aerleon/lib/proxmox.py
```diff
@@ -0,0 +1,658 @@
...
+    @staticmethod
+    def _singlePortFmt(port: int) -> str:
+        return str(port)
+
+    @staticmethod
+    def _portTupleFmt(port: tuple) -> str:
+        if port[0] == port[1]:
+            return str(port[0])
+        else:
+            return '%d-%d' % port
```
ankenyr, line 202: nit: Prefer f strings where possible.

aerleon/lib/proxmox.py
```diff
@@ -0,0 +1,658 @@
...
+
+        # proxmox firewall supports multiple sources/destinations per rule
+        if source_interface:
+            options.append("-iface %s" % source_interface)
+
+        # we cannot use address tokens since IPsets are defined in separate files (at cluster
+        # or at host level) and we're only outputting one file at a time, which may not be
+        # cluster/host firewall configuration.
+        if destination_addresses:
+            options.append("-dest %s" % ','.join(map(to_network_addr, destination_addresses)))
```
ankenyr, line 458: prefer f strings here, they are generally faster and lead to more readable code.

aerleon/lib/proxmox.py
```diff
@@ -0,0 +1,658 @@
...
+            options.append(
+                "-dport %s" % ','.join(map(lambda p: str(ProxmoxPort(p)), destination_ports))
+            )
+
+        if protocol in ProxmoxIcmp.ICMP_PROTOS:
+            options.append('-icmp-type %s' % str(ProxmoxIcmp(protocol, icmp_type, icmp_code)))
+
+        if logging:
+            logging_key = 'true'
+            log_option = next(
```
ankenyr, line 478: This might be better broken into a small method.

aerleon/lib/proxmox.py
```diff
@@ -0,0 +1,658 @@
...
+        return list(set(list_a + list_b))
+
+    def merge(self, other):
+        self_keys = list(self.keys())
+        other_keys = list(other.keys())
+        ret = dict()
+        for k in self_keys + other_keys:
+            if k in self_keys and k in other_keys:
+                if isinstance(self[k], list) and isinstance(other[k], list):
+                    ret[k] = self._merge_list(self[k], other[k])
```
ankenyr, line 67: Should the lists be sorted after they are merged?

aerleon/lib/proxmox.py
```diff
@@ -0,0 +1,658 @@
...
+
+
+class Proxmox(aclgenerator.ACLGenerator):
+    """Proxmox firewall policy object"""
+
+    # aerleon class props
+    _PLATFORM = 'proxmox'
+    SUFFIX = '.prxmxfw'
+    _TERM = Term
+    _LOG_LEVELS = list(_TERM.LOG_LEVELS_MAP_OPTIONS.values())
```
ankenyr, line 500: Any reason some of the constants are defined in Term when they are only used in Proxmox?

aerleon/lib/proxmox.py
```diff
@@ -0,0 +1,658 @@
...
+            for t in self.filter_options[2:]:
+                ingested_options = []
+                for o in available_zone_options:
+                    ingested_options.append(o.ingest(t))
+                if not any(ingested_options):
+                    raise UnsupportedFilterOptionError("incorrect filter option directive %s" % t)
+            incomplete_options = list(filter(lambda o: not o.complete(), available_zone_options))
+            if incomplete_options:
+                raise UnsupportedFilterOptionError(
+                    "missing or incorrect value for filter option(s) %s", incomplete_options
```
ankenyr, line 629: Can incomplete_options ever be an object? If so then we should probably wrap this in str.

ankenyr (commented):
Looks great so far. Do you plan to add more tests? How is the coverage at this point?

aerleon/lib/proxmox.py
```diff
@@ -0,0 +1,658 @@
...
+
+    def merge(self, other):
+        self_keys = list(self.keys())
+        other_keys = list(other.keys())
+        ret = dict()
+        for k in self_keys + other_keys:
+            if k in self_keys and k in other_keys:
+                if isinstance(self[k], list) and isinstance(other[k], list):
+                    ret[k] = self._merge_list(self[k], other[k])
+                else:
```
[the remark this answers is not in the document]
loulecrivain: don't think there'd be cases where we wouldn't merge, data validation is already handled through child classes at this point (or there is no data at all, so it'd be a no-op)

loulecrivain:
Code coverage is at ~76%

I'll add more tests for mixed-direction policies and proxmox target options and we should be good

loulecrivain:
coverage is now at 94%, hth

ankenyr (approved):

ankenyr:
@all-contributors please add @loulecrivain  for code and documentation

PR #406, opened 2025-06-25
Add documentation for AclCheck command line and API

jtwb:
This change adds basic documentation for the AclCheck command line tool and API. It includes a test case for the example included in the docs page.


jtwb:
> [nitpick] The expected summary message includes specific indentation which might make the test brittle if output formatting is adjusted. Consider normalizing whitespace in the summary message or documenting the expected format more explicitly.

That is the actual output format, odd as it is.

ankenyr (approved):

PR #409, opened 2025-08-12
Generate API: add includes, include_path support

jtwb:
**New Features**

Adds support for includes in policies processed by the `Generate` API through new optional parameters: `include_path` and `includes`.
* If `include_path` is provided, `include` directives in user policies may reference YAML files within include_path. A strict subpath check is performed on user `include` directives to prevent access to files outside of include_path via symlinks or relative paths. Additionally only files ending in .yaml or .yml may be used as include targets.
* If `includes` is provided (a dictionary of names mapping to includable PolicyDicts), user `include` directives will refer to values in the `includes` dictionary only. No filesystem access is performed with this option.
* If neither option is given, the `Generate` API maintains its current behavior which is to raise an error when processing a policy with an `include` directive.
* `include_path` and `includes` are mutually exclusive.

### DIFF
--- a/aerleon/api.py
+++ b/aerleon/api.py
@@ -38,7 +38,7 @@ have a single filter, which must have a "header" section and a "terms" list. The
 "header" instructs Aerleon to produce Cisco ACL output. The "terms" list defines
 the access control behavior we want for this filter.
 
-```
+```python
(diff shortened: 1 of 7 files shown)


ankenyr:
This is looking fine to me

PR #418, opened 2025-08-27
Fix #417 - aclcheck example_policy documentation

### LINKED ISSUE #417: `aclgen` example gives warning
The `aclcheck` [example in the documentation](https://aerleon.readthedocs.io/en/latest/aclcheck_api/) produces a warning when run:

```
WARNING:absl:Unexpected term keyword kvs
  Filter: test-filter
          term: allow-web-traffic
                accept
```

ABruihler:
This fixes the example causing a warning in #417, updates the other related documentation, and changes the tests so they would catch similar generated `logging` warnings in the future.

### DIFF
--- a/aerleon/api.py
+++ b/aerleon/api.py
@@ -47,9 +47,7 @@ cisco_example_policy = {
                 "targets": {
                     "cisco": "test-filter"
                 },
-                "kvs": {
-                    "comment": "Sample comment"
-                },
+                "comment": "Sample comment",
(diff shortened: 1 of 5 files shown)


ABruihler:
From looking further it seems like `tests/api/api_test.py` also has this `kvs` format, and I don't yet understand why I am having issues but it seems like it is succeeding on the repo GitHub Actions.

PR #428, opened 2025-09-23
f-string reformats using flynt

ABruihler:
This PR uses [`flynt`](https://pypi.org/project/flynt/) to batch reformat many "%-formatted" and .format(...) strings into Python 3.6+'s "f-strings".  As the Flynt docs say:

> Not only are they more readable, more concise, and less prone to error than other ways of formatting, but they are also faster!

Flynt does not guarantee that all formatted strings will be the same as before conversion, but this should be pretty safe.  I have not read through every change that this includes, but all tests pass and on skimming through things look good.  Also, this PR does not use the Flynt `--aggressive` flag that includes additional conversions that potentially change behavior.

### DIFF
--- a/aerleon/__main__.py
+++ b/aerleon/__main__.py
@@ -7,5 +7,5 @@ try:
     aclgen.EntryPoint()
     rc = 0
 except Exception as e:
-    print('Error: %s' % e, file=sys.stderr)
+    print(f'Error: {e}', file=sys.stderr)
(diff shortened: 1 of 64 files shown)


ankenyr:
This looks good, thanks for cleaning up the comments in aclgenerator.py and aclgenerator_test.py that are no longer needed.

PR #447, opened 2025-12-15
pyupgrade 3.10+ syntax updates & GitHub Actions check

ABruihler:
This PR adds Python 3.10+ syntax using pyupgrade, now that Python 3.9 is unsupported.  

It also adds a GitHub Actions step validate/format to check for Python 3.10+ syntax using pyupgrade.


ABruihler:
As shown in the Checks history of this PR, the validate/format GitHub Action will now catch when python files were edited but do not meet the 3.10+ pyupgrade syntax and fail the validate/format step.

ankenyr (approved):
looks great! Thank you!

Issue #454, opened 2026-01-16
`aclcheck` type hints are unnecessarily narrow to `str`


ABruihler:
The signature for `api.AclCheck` is currently:
```python3
def AclCheck(
    input_policy: policy_builder.PolicyDict,
    definitions: naming.Naming,
    src: str | None = None,
    dst: str | None = None,
    sport: str | None = None,
    dport: str | None = None,
    proto: str | None = None,
):
```

These fields seem to be unnecessarily narrowly scoped to `str` when they can actually take e.g. an `IPv4Address` or integer etc.  

It would be helpful for this function to have better type hinting.  It looks like related type hints beyond the `api.AclCheck` signature could also be improved - like on `aclcheck.AclCheck`, `nacaddr`, `port`, etc.

PR #455, opened 2026-01-17
Fix `aclgen` type hinting

ABruihler:
This PR addresses #454 by adding type hinting to `api.AclCheck` and certain deeper related functions.

### DIFF
--- a/aerleon/lib/aclcheck.py
+++ b/aerleon/lib/aclcheck.py
@@ -81,11 +81,11 @@ class AclCheck:
         cls,
         policy_dict: policy_builder.PolicyDict,
         definitions: naming.Naming,
-        src: IPv4Address | IPv6Address | IPv4Network | IPv6Network | str | Literal["any"] | None,
-        dst: IPv4Address | IPv6Address | IPv4Network | IPv6Network | str | Literal["any"] | None,
-        sport: int | str | Literal["any"] | None,
-        dport: int | str | Literal["any"] | None,
-        proto: str | Literal["any"] | None,
+        src: IPv4Address | IPv6Address | IPv4Network | IPv6Network | str | Literal["any"],
+        dst: IPv4Address | IPv6Address | IPv4Network | IPv6Network | str | Literal["any"],
+        sport: int | str | Literal["any"],
+        dport: int | str | Literal["any"],
+        proto: str | Literal["any"],
     ) -> Self:
         """Construct an AclCheck object from a PolicyDict + Naming object."""
         policy_obj = policy.FromBuilder(policy_builder.PolicyBuilder(policy_dict, definitions))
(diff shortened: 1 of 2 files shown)


aerleon/lib/port.py
```diff
@@ -149,7 +149,7 @@ def __eq__(self, other):
             raise NotSinglePort('Comparisons cannot be performed on port ranges')
 
 
-def Port(port):
+def Port(port: int | str) -> int:
```
ABruihler: The `int` type hints here are further constrained - i.e. `0 <= port <= 65535`.  There are a few ways of representing this in Python like `typing.Annotated[int, "0 <= port <= 65535"]` or using Pydantic, but those seem to be bigger style choices to make so I didn't just pick something for this case.

aerleon/lib/aclcheck.py
```diff
@@ -62,32 +66,51 @@ class AclCheck:
...
-    ):
+        src: (
+            IPv4Address | IPv6Address | IPv4Network | IPv6Network | str | Literal["any"] | None
+        ) = 'any',
+        dst: (
+            IPv4Address | IPv6Address | IPv4Network | IPv6Network | str | Literal["any"] | None
+        ) = 'any',
+        sport: int | str | Literal["any"] | None = 'any',
+        dport: int | str | Literal["any"] | None = 'any',
+        proto: str | Literal["any"] | None = 'any',
```
ABruihler: Do we want to allow `None` here, or exclude it from the type hints to discourage passing it in?  

It seems like the cleanest thing would be to not type hint `None` as allowed here and expect the caller to not pass the argument and rely on the default value in such situations, but that would add some additional complexity elsewhere.

aerleon/lib/aclcheck.py
```diff
@@ -62,32 +66,51 @@ class AclCheck:
...
 
+    pol_object: policy.Policy
+    pol: policy.Policy
+    src: nacaddr.IPv4 | nacaddr.IPv6 | Literal["any"]
+    dst: nacaddr.IPv4 | nacaddr.IPv6 | Literal["any"]
+    sport: int | Literal["any"]
+    dport: int | Literal["any"]
+    proto: str | Literal["any"]
+    matches: list
+    exact_matches: list
```
ABruihler: It would be nice to type hint these further but I didn't investigate them as part of this PR.

aerleon/lib/aclcheck.py
jtwb: I'm late to the party here, but as this interface is long-published and there could exist working user code that passes `None` (with the exception of `proto` where I think you found and fixed a bug), we should take care to accommodate that program. Or, deprecate, and be loud and clear about when and why a deprecation is occurring. I'd lean towards continuing to allow `None` on account of laziness as the cost to keep the existing behavior isn't high.

Issue #453, opened 2026-01-16
`AclCheck` zone support


ABruihler:
Currently, it seems that `AclCheck` ignores zones while evaluating policy, for example with `source-zone` and `destination-zone` on specific policy terms for JuniperSRX global policies (i.e. `targets: srx: from-zone all to-zone all inet`).  This means that `AclCheck` results can be incorrect for such policies including zones on terms.

Related, policies that are specific zone-to-zone policies (i.e. not global `from-zone all to-zone all inet` for JuniperSRX) possibly could/should have zones handled differently by `AclCheck`.

PR #456, opened 2026-01-18
Adding zone checking in aclcheck.

ankenyr:
Zones, both source and destination, can now be specified in aclcheck.

When a zone (source or destination)  is added, similar to ip or port, it will check the terms. If the term does not specify the zones allowed it is assume to be "any". If the zones do not match, the term is considered to not be matched. If the zone is any, then all zone names match. 

This does not take into consideration header options and only considers information in the terms. This could be improved in the future but it well beyond the scope of this issue.

@ABruihler for visibility

https://github.com/aerleon/aerleon/issues/453

### DIFF
--- a/aerleon/lib/aclcheck.py
+++ b/aerleon/lib/aclcheck.py
@@ -76,6 +76,8 @@ class AclCheck:
     proto: str | Literal["any"]
     matches: list
     exact_matches: list
+    source_zone: str | Literal["any"]
+    destination_zone: str | Literal["any"]
 
     @classmethod
     def FromPolicyDict(


aerleon/lib/aclcheck.py
```diff
@@ -118,6 +120,16 @@ def __init__(
                 self.dst = nacaddr.IP(dst)
             except ValueError:
                 raise AddressError(f'bad destination address: {dst}\n')
+        # validate zones
```
ABruihler:
```suggestion

        # validate zones
```
consistent section newline breaks

aerleon/lib/aclcheck.py
```diff
@@ -137,6 +149,19 @@ def __init__(
                     logging.debug('dstaddr does not match')
                     continue
                 logging.debug('dstaddr matches: %s', self.dst)
+                # source-zone matching if requested. If the term does not specify
+                # a source_zone, treat it as 'any' (match all zones).
+                if self._ZoneMatch(self.source_zone, term.source_zone) is False:
```
ABruihler:
```suggestion
                if not self._ZoneMatch(self.source_zone, term.source_zone):
```

aerleon/lib/aclcheck.py
```diff
@@ -256,6 +281,19 @@ def _PossibleMatch(self, term):
...
         return ret_str
 
+    def _ZoneMatch(self, zone, term_zone) -> bool:
+        """Check if zone matches term zone.
+
+        Args:
+          zone: A string for the zone to check
+          term_zone: A list of zones from the term
+        """
+        if term_zone == [] or zone == 'any':
```
ABruihler:
```suggestion
          term_zone: A collection of zones from the term
        """
        if not term_zone or zone == 'any':
```

If something for `term_zone` other than a `list` is provided (such as a `set`), `not term_zone` can handle it better.

ABruihler (commented):
Added a few suggestions, mostly for adding type hints.

ankenyr:
Thanks for the help. @ABruihler

jtwb (approved):
Looks good. I think the release notes should reflect that this feature applies to ACL terms which use source-zone and destination-zone, not zone-to-zone policies.

PR #471, opened 2026-02-12
Improve policy_term.py type hints

ABruihler:
This PR improves some type hints in policy_term.py:

* sets type hints directly instead of quoting.
* converts invariant `list[int | str]` hints to covariant `Sequence[int | str]` hints

I started this branch after noticing a pyright type error in our own project about the invariant/covariant issue:
```
/Users/me/file.py
  /Users/me/file.py:123:45 - error: Could not assign item in TypedDict
    Type "list[str]" is not assignable to type "int | str | list[int | str]"
      "list[str]" is not assignable to "int"
      "list[str]" is not assignable to "str"
      "list[str]" is not assignable to "list[int | str]"
        Type parameter "_T@list" is invariant, but "str" is not the same as "int | str"
        Consider switching from "list" to "Sequence" which is covariant (reportGeneralTypeIssues)
```

### DIFF
--- a/aerleon/lib/policy_builder.py
+++ b/aerleon/lib/policy_builder.py
@@ -3,6 +3,7 @@
 import enum
 import sys
 import typing
+from collections.abc import Sequence
 from dataclasses import dataclass, field
 from datetime import date
 from typing import Annotated, TypeAlias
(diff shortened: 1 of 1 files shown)


jtwb:
@ABruihler This PR needs to be updated to include newer commits & resolve conflicts.

ABruihler:
Huh, I must have gotten my branches crossed somehow.  This was sorted out in #475

PR #475, opened 2026-02-24
Improve policy_builder.py typing

ABruihler:
This PR improves type hints in policy_builder.py:
* don't quote type hints
* use covariant `Sequence` instead of `list` when allowing multiple content types
* don't import all of `typing`
* other misc type hint improvements

### DIFF
--- a/aerleon/lib/policy_builder.py
+++ b/aerleon/lib/policy_builder.py
@@ -669,7 +669,7 @@ class _Builtin:
         return cls(keyname, *cls.BUILTINS[keyname])
 
     @property
-    def recognizer(self) -> "TValue | TComposition":
+    def recognizer(self) -> TValue | TComposition:
         """The recognizer specific to this Builtin instance."""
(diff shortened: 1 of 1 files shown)


aerleon/lib/policy_builder.py
```diff
@@ -40,42 +42,37 @@
 else:
     from typing import TypedDict
 
-if typing.TYPE_CHECKING:
```
jtwb, line 43: `aerleon.lib.naming` is a fairly large import and used for type checking only, so performance best practice calls for its conditional import.

jtwb (approved):
Approved but please see the conditional import comment.

aerleon/lib/policy_builder.py
ABruihler: Done - `from aerleon.lib.naming import Naming` is now conditional on `typing.TYPE_CHECKING` again.

PR #484, opened 2026-05-07
NokiaSRL: support named prefix-lists in ACL entries

### LINKED ISSUE #483: Nokia SR-Linux: Support matching on named prefix-lists
## Issue

On our Juniper platforms we have some named prefix-lists that are dynamically generated by our automation, which we can then include in our Aerleon defined policies using the _source-prefix_ and _destination-prefix_ tokens.  An example term we have is:
```
      - name: allow_bgp_src
        source-prefix: bgp-sessions bgp-sessions-overlay
        protocol: tcp
        source-port: BGP
        action: accept
```

Unfortunately, while the Nokia SR-Linux platform supports a similar use of named-prefix lists (defined under _"acl match-list"_ in the config), Aerleon currently does not provide a way to reference these in policies.

## Fix

I've been trying to hack something together to support this for our own use-case.  I used the _source-prefix_ and _destination-prefix_ token names to try to match what is done for other vendors (Juniper, Arista etc.)  Right now it all works, and for instance with the above input generates the below SR Linux ACL entry:

```json
{
  "action": {
    "accept": {}
  },
  "description": "allow_bgp_src",
  "match": {
    "ipv4": {
      "protocol": 6,
      "source-ip": {
        "prefix-list": [
          {"name": "bgp-sessions"},
          {"name": "bgp-sessions-overlay"}
        ]
      }
    },
    "transport": {
      "source-port": {"value": 179}
    }
  },
  "sequence-id": 5
}
```

I'll put it together in a PR anyway and try to get some feedback if the approach taken makes sense.

topranks:
Closes #483

### Summary

Adds support for `source-prefix` and `destination-prefix` tokens in the `nokiasrl` generator, so users can reference existing named prefix-lists defined in the configuration.  The tokens are already used elsewhere (juniper, arista_tp, iptables) with similar semantics.

### Example

With this patch a term like this:
```yaml
- name: allow_bgp_src
  source-prefix: bgp-sessions bgp-sessions-overlay
  protocol: tcp
  source-port: BGP
  action: accept
```
Will produce an SR-Linux ACL entry like this:

```json
{
  "action": {"accept": {}},
  "description": "allow_bgp_src",
  "match": {
    "ipv4": {
      "protocol": 6,
      "source-ip": {
        "prefix-list": [
          {"name": "bgp-sessions"},
          {"name": "bgp-sessions-overlay"}
        ]
      }
    },
    "transport": {
      "source-port": {"value": 179}
    }
  },
  "sequence-id": 5
}
```

## Notes for review

- I re-used the existing `source-prefix` / `destination-prefix` names rather than creating new Nokia-specific tokens, as the concept here is the same as the other places those names are used.

- Combining `source-prefix` and `source-address` on the same term raises an exception.  SR Linux YANG doesn't allow both in the same `source-ip` container so this seemed sensible.

### Changes

- `aerleon/lib/nokiasrl.py` — extends `IPPrefix` to accept either a literal `prefix` or a `prefix-list` array, adds `SetSourcePrefixList` / `SetDestPrefixList` setters, adds `source_prefix` / `destination_prefix` to the supported token set, and raises `PrefixListWithAddressError` when a term combines `source-prefix` with `source-address` (same for destination).
- `tests/regression/nokiasrl/nokiasrl_test.py` — five new tests: single source / destination prefix-list, multiple source prefix-lists, and the two conflict-detection error paths.
- `docs/reference/generators.md` — documents the two new tokens in the NokiaSRL section.

### Testing

- `poetry run pytest tests/regression/nokiasrl/` — 24 tests pass (19 existing + 5 new).

This is my first contribution to Aerleon, so any feedback on style or approach is very welcome.

### DIFF
--- a/aerleon/lib/nokiasrl.py
+++ b/aerleon/lib/nokiasrl.py
@@ -27,8 +27,18 @@ from aerleon.lib.policy import Term
 R24_3_2 = "r24.3.2"  # Option flag to generate release >= 24.3.2 syntax
 
 
-class IPPrefix(TypedDict):
-    prefix: str
+class PrefixListRef(TypedDict):
+    name: str
+
+
+IPPrefix = TypedDict(
+    "IPPrefix",
+    {
+        "prefix": str,
+        "prefix-list": list[PrefixListRef],
+    },
+    total=False,
+)
 
 
 class PortRange(TypedDict):
@@ -103,6 +113,10 @@ class EstablishedWithNonTcpUdpError(Error):
 
 
 class UnsupportedLogging(Error):
+    pass
+
+
+class PrefixListWithAddressError(Error):
     pass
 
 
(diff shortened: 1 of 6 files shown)


XioNoX:
@ankenyr @jtwb2 that change would be useful for us, I was wondering if you could have a look? Thanks

PR #485, opened 2026-05-23
Remove SyncManager from single-renderer code paths

jtwb:
Currently, a SyncManager instance is created for every call to api.Generate and aclgen.Run . SyncManager is a resource-intensive object and creating it here can interfere with calling the Generate API from within multi-threaded environments or web servers. It's also not needed for the Generate API at all, as the Generate API always runs with max_renderers=1 .

This PR just instantiates the SyncManager only in cases when parallelism is activated via aclgen.Run with max_renderers > 1.

PR also includes a Docker-based repro script.

A recent user report states:

> has anyone seen any weird "BlockingIOError: [Errno 11] Resource temporarily unavailable" when running aerleon from another thread in python 3.13? I asked claude about it and it thinks in python 3.13 SyncManager should be changed to ProcessPoolExecutor to avoid some race condition


ankenyr (commented):
Made a few comments but lgtm otherwise.

ankenyr (approved):

PR #489, opened 2026-06-04
Nokia SRL: add support for restrict-address-family

XioNoX:
I noticed it when pushing [this](https://gerrit.wikimedia.org/r/c/operations/homer/public/+/1296933/1/policies/srl-common-loopback.yaml) change to the target `nokiasrl: cpm mixed stats r24.3.2` :
```
      - name: allow_traceroute
        comment: T348120
        source-address: RFC1918_10
        destination-port: TRACEROUTE
        protocol: udp
```

As `RFC1918_10` is only a v4 prefix, instead of not creating a v6 term (which would be safer), it created a v6 term with no `source-address` (diff slightly modified for readability), so "wide open":
```diff
          acl-filter cpm type ipv6 {
+             entry 445 {
+                 description allow_traceroute
+                 match {
+                     ipv6 {
+                         next-header 17
+                     }
+                     transport {
+                         destination-port {
+                             range {
+                                 start 33434
+                                 end 33534
+                             }
+                         }
+                     }
+                 }
+                 action {
+                     accept {
+                     }
+                 }
+             }
          }
```

Adding `restrict-address-family` allows me to solve (or workaround) this issue, and it's a term present in other platforms too (cisco, juniper).

### DIFF
--- a/aerleon/lib/nokiasrl.py
+++ b/aerleon/lib/nokiasrl.py
@@ -233,7 +233,13 @@ class NokiaSRLinux(openconfig.OpenConfig):
         """
         supported_tokens, supported_sub_tokens = super()._BuildTokens()
 
-        supported_tokens -= {'platform', 'platform_exclude', 'verbatim', 'icmp-type'}
+        supported_tokens -= {
+            'platform',
+            'platform_exclude',
+            'verbatim',
+            'icmp-type',
+            'restrict_address_family',
+        }
 
         supported_sub_tokens['action'] = {'accept', 'deny'}  # excludes 'reject'
         supported_sub_tokens['option'] = {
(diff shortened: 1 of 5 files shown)


ankenyr:
Mind adding a description into here? To me this sound like it should be the default behavior.

XioNoX:
@ankenyr Yeah I'd tend to agree but I didn't want to start looking at changing the default behavior.

Description updated for context, let me know if something needs to be clarified.

XioNoX:
@ankenyr do you need any other information from me ? Thanks

ankenyr:
Thanks for the poke bringing this back to my attenion
https://github.com/aerleon/aerleon/blob/7307a8875acd6e272aff9f623c65af02f748feee/docs/reference/generator_patterns.md?plain=1#L69
I believe we should change this to adhere to the general pattern. By default if one or both sides of the ips are filtered out, it should not be rendered. The intent for the rule says that traffic should be allowed between x and y. If there is no x or not y after filtering, there is nothing to allow.
@jtwb for visibility.

XioNoX:
@ankenyr I'm wondering if I should abandon this pull request,
On one hand the default behavior should make this change less useful, but I don't think I know how to change the default behavior.
On the other hand, `restrict-address-family` is something that's supported by other platforms, so it might be useful regardless of the default behavior.

Thanks

PR #491, opened 2026-07-21
fortigate: support source-fqdn / destination-fqdn

jordanharrison00:
## Summary

FortiGate configs commonly use FQDN address objects (`config firewall address`
with `set type fqdn`), but the FortiGate generator rejects any term using
`source-fqdn`/`destination-fqdn`:

```
<term> contains unsupported keywords (destination_fqdn) for target fortigate
```

This adds FQDN support to the FortiGate generator.

## Root cause

FQDN is opt-in per generator — `ACLGenerator._BuildTokens()` does not include
`source_fqdn`/`destination_fqdn`. The Junos SRX generator advertises them and
renders `dns-name`; the FortiGate generator never advertised them and had no FQDN
rendering, so keyword validation rejects such terms.

## Change

FortiGate models an FQDN as an address object of `type fqdn` referenced from a
policy's `srcaddr`/`dstaddr` (there is no separate FQDN policy field, unlike Junos
`dns-name`):

- Advertise `source_fqdn`/`destination_fqdn` in `Fortigate._BuildTokens`.
- Add `FortigateFqdnAddress`, which emits `set type fqdn` / `set fqdn "<value>"`.
- `Term._TranslateFqdns` registers the FQDN address objects (a naming object
  holding multiple FQDNs expands to `NAME_0`/`NAME_1`) and folds their tokens into
  the v4 `srcaddr`/`dstaddr`.
- Emit the FQDN objects in the `config firewall address` section.

## Example

```yaml
filters:
  - header:
      targets:
        fortigate: from-zone port1 to-zone port2
    terms:
      - name: to-fqdn
        destination-fqdn: GOOGLE_DNS
        action: accept
```

with `GOOGLE_DNS` defined as `fqdn: dns.google.com` produces:

```text
config firewall address
    edit GOOGLE_DNS
        set type fqdn
        set fqdn "dns.google.com"
    next
end
...
        set dstaddr "GOOGLE_DNS"
```

## Tests

Adds `testDestinationFqdn`, `testSourceFqdn`, and
`testMultipleFqdnValuesExpandToIndexedObjects` to
`tests/regression/fortigate/fortigate_test.py`; the full FortiGate regression
suite passes.

## Known Rough Edges

- A single naming object carrying both an IP and an FQDN, referenced on the same
  side via both `destination-address` and `destination-fqdn`, would collide
  (`addrgrp` vs `address` with the same name); a follow-up could suffix the FQDN
  object to disambiguate.

### DIFF
--- a/tests/regression/fortigate/fortigate_test.py
+++ b/tests/regression/fortigate/fortigate_test.py
@@ -428,6 +428,20 @@ class FortigateTest(parameterized.TestCase):
         # Different IPs from MIXED_P for exclusion testing
         self.naming._ParseLine('MIXED_Q = 1.1.1.3/32 2001:db8::3/128', 'networks')
 
+        # FQDN definitions (only expressible via the object/YAML naming path).
+        self.naming.ParseDefinitionsObject(
+            {
+                'networks': {
+                    'GOOGLE_DNS': {'values': [{'fqdn': 'dns.google.com'}]},
+                    'MULTI_FQDN': {
+                        'values': [{'fqdn': 'a.example.com'}, {'fqdn': 'b.example.com'}]
+                    },
+                },
+                'services': {},
+            },
+            '',
+        )
+
     @capture.stdout
     def testNoAddressesInet6OutputsAll(self):
         pol = policy.ParsePolicy(INET6_HEADER + ALL_IPS, self.naming)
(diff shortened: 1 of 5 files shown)


tests/regression/fortigate/fortigate_test.py
```diff
@@ -466,6 +480,51 @@ def testInet(self):
         self.assertNotIn('dstaddr6 ', str(acl))
         print(acl)
 
+    def testDestinationFqdn(self):
```
ankenyr, line 484: Is there a reason you did not use the @capture.stdout decorator here? This captures a render and comapres it to a golden copy named after the test in https://github.com/aerleon/aerleon/tree/main/tests/regression/fortigate

jordanharrison00: There's most certainly a reason! That reason is **ignorance**.  I have updated the PR with appropriate tests that are now in-line with the other tests present.

ankenyr:
Did you need to update any of the docs?
docs/reference/generators.md

ankenyr:
Have you tested these with the other headers like inet6, mixed, or local-in-policy? Would this fail open with an inet6 policy given these are ipv4 only objects?

